Issue #29, opened 2014-12-17
CUBA keywords need probably to be grouped`


itziakos:
It looks that some CUBA keywords can only have meaning as part of a specific component (e.g. 
keywords `BOND_LABEL` and  `BOND_TYPE` probably only make sense for a Particle).

Having one `DataContainer` type might be convenient now but having all the keywords supported when they are not necessary is performance-wise inefficient and makes bad use of memory and hard-disk space. We should also consider that it can lead to confusions (e.g. a mesh cell with `SIMULATION_DOMAIN_DIMENSIONS` does sound a little weird).

We should probably use a subtype of the DataContainer for the different components where such an object is used.  The implementation could be as simple as binding each DataContainer subtype into a specific CUBA keyword group.

Nevertheless, I believe that this avenue needs to be discussed after the current first phase of the api implementation is delivered.

itziakos:
The plan for this feature is to do the following:

- Each entry of the cuba.yml will be augmented with `data` attribute that will have value true/false based on the fact that needs to be supported by a DataContainer
- Other containers e.g. configuration containers will have their own attribute e.g. `config` that will also be true/false with similar meaning. A keyword can belong to more than one enum description
- The cuba-generate then will be able to produce a cuba.py containing two separate enum classes containing the desired keywords.
- For each enum class we can create a specialized dictionary like container that will support only the keywords provided by the associated enum.

Please note that since the int associated with each keyword is going to has the same value for both enums coping common key,value pairs between the different containers will be supported without any problem.

ahashibon:
The most straightforward way is to add to which SimPhoNy component(s) (SD, CM, SP, BC) each CUDS keyword may be used in. This is in addition to the information we have now. Any additional grouping would not be very useful at the moment, we can reconsider this later on when we consolidate a more complete list at the end of Y2.

itziakos:
It sounds reasonable, thanks @ahashibon.

stefanoborini:
Closing as it was for discussion and is no longer relevant.

PR #31, opened 2014-12-19
io support for data container

itziakos:
This a  PR contains the io support for the data container items

In more detail:
- `DataContainer`s are saved in a hdf5 table node which contains a nested column for the key/data values and a column masking out the invalid keywords. The records are mapped using uuids.
- The list of the CUBA keywords is now provided in yaml format so that it is easy to be edited by users and at the same time can be easily loaded into scripts for code generation.
- A code generator script is (i.e. cuba-generate) provided to generate the `CUBA` enum class used in `DataContainer` and the `Data` table description used in the new `DataContainerTable` class.

**Pending**:
- [x] Finish the DataContainerTable
- [x] Update the setup.py and include any new requirements
- [x] Provide a benchmarking script for the DataContainerTable.
- [x] Cleanup

**Notes**:
- Missing CUBA keywords were added, but the CUBA keyword correspondence to integers has changed.


itziakos:
@dpinte, @roigcarlo, @nathanfranklin @SGGgarcia This PR is ready for a first review. I will be updating the pending components shortly

itziakos:
@roigcarlo do you have time available to review this PR?

roigcarlo:
@itziakos I'll start asap

itziakos:
Thanks @roigcarlo

simphony/io/data_container_table.py
```diff
@@ -0,0 +1,169 @@
...
+                data_row[positions[key]] = data[key]
+                mask[positions[key]] = True
+            row['mask'] = mask
+            row['Data'] = tuple(data_row)
+            row.update()
+            # see https://github.com/PyTables/PyTables/issues/11
+            row._flush_mod_rows()
+            break
+        else:
+            raise KeyError('Index {} is not found'.format(uid))
```
itziakos: I think that raising an error here is incorrect since the code
```python
uid = uuid.uuid4()
data_container_table[uid] = DataContainer()
```
should add an empty row in the table mapped to uid instead of raising and error. This will make using the DataContainerTable inside the CUDS container much more convenient.

itziakos: It is also the expected behaviour for the full Mutable-Mapping api

itziakos:
@roigcarlo, I have updated (see comments https://github.com/simphony/simphony-common/pull/31#discussion_r23304611) and cleaned-up a little the code

roigcarlo:
@itziakos I agree with https://github.com/simphony/simphony-common/pull/31#discussion_r23304736 and it seems the logical behaviour.

simphony/scripts/cuba_generate.py
```diff
@@ -0,0 +1,99 @@
+import click
+import yaml
+
+# Cuba keywords that are excludes from DataContainers
+CUBA_DATA_CONTAINER_EXLCUDE = ['Id', 'Position']
```
roigcarlo: Couldn't be this an attribute in the yml? All other properties like shape, key,... are there,

itziakos: Good point. We could have an attribute `data_container` that has a true or false value based on the fact that it is one of the DataContainer supported keywords. Other containers could also have their own attribute.

This is related with https://github.com/simphony/simphony-common/issues/29 where we will have groups off keywords which various dictionary like container will support (see discussion in https://github.com/simphony/simphony-lammps-md/pull/6#discussion_r23318337). 

While I believe that you are right and we should implement the new attribute in yaml, I would prefer to implement this improvment in a separate PR, what do you think.

roigcarlo: Ok, sounds good to me

simphony/core/cuba.yml
```diff
@@ -0,0 +1,476 @@
...
+  shape: [3]
+  type: double
+- description: Length in angstroms of the shape of the materials
+  domain: [VIS]
+  key: SHAPE_LENGTH
+  name: ShapeLength
+  number: 11
+  shape: [3]
+  type: double
+- description: Radio for a spherical material
```
roigcarlo: Typo: Radio -> Radius?

simphony/scripts/cuba_generate.py
itziakos: Thanks, updated issue #29 to describe changes

simphony/core/cuba.yml
itziakos: Done

itziakos:
@roigcarlo is this good to merge?

roigcarlo:
I didn't detect any other problem, so yes, for my part all good!

itziakos:
Thanks for the review @roigcarlo

PR #33, opened 2015-01-21
use uuid for unique id for particle io

itziakos:


itziakos:
@nathanfranklin, @roigcarlo and @tuopuu, This PR augments the `H5Particle` hdf5 backed container to support using the uuid and saving the `data` information of the `Particles` and `Bonds`. I have also implemented an example on how we can create template objects to simplify writing tests for the high level CUDS containers. Using these templates, both the `Particles` and `H5Particles` containers are evaluated against the same tests and we can make sure that they will behave the same (including having the same errors). I also did some minor cleanup and updated the raised errors to follow better the commonly used patterns in python. In short `KeyError` is raise when the function is getting a `key` objects that maps to an item (e.g. get_xxx, remove_xxx). while `ValueError` is raised when the item is passed as part of signature. Finally, I factored out a convenience abstract class that implements an augmented dictionary proxy to an HDF5 backed tables for item and data.(Test for this class will arrive shortly)

nathanfranklin:
> I have also implemented an example on how we can create template objects to simplify writing tests for the high level CUDS containers.

Thanks.  This looks very good.

>  In short KeyError is raise when the function is getting a key objects that maps to an item (e.g. get_xxx, remove_xxx). while ValueError is raised when the item is passed as part of signature.

Will the get_xxx, remove_xx of H5CUDS also have to raise then KeyError instead of ValueError.

itziakos:
> Will the get_xxx, remove_xx of H5CUDS also have to raise then KeyError instead of ValueError.

Yes, we should probably update that also.

itziakos:
Tests added. And I just released that IndexedDataContainerTable, DataContainerTable and H5CUDSItems could be redesigned a little more elegantly and probably perfoming fewer lookups, but I will that for a separate optimization/cleanup task.

itziakos:
Finally, this PR is ready for review

nathanfranklin:
LGTM.

maybe @roigcarlo or @tuopuu could also take a look.

itziakos:
Please do not merge until #71  is merged

PR #34, opened 2015-01-21
Use coveralls instead of reporting on travis-ci

itziakos:
This PR add the use of coveralls-io to report coverage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,10 @@ install:
   - pip install numexpr cython==0.20
   - pip install -r requirements.txt
   - pip install -r dev_requirements.txt
+  - pip install coveralls
   - python setup.py develop
 script:
   - flake8 .
   - coverage run -m unittest discover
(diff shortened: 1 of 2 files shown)


itziakos:
@nathanfranklin  and @ahashibon this PR adds the coveralls service that show detailed coverage reports for the travis-ci builts. It will also sent comments to the PR when the coverage changes due to the code in the PR. In that sense it is easy to spot PRs that degrade the test coverage of the library so we can take action to rectify.

itziakos:
The information is freely available (like the built results of travis-ci for opensource project) at https://coveralls.io/r/simphony/simphony-common

itziakos:
Please note that our current coverage is around 88% which is a valid statement coverage. For more on coverage metric one can read http://www.bullseye.com/coverage.html

nathanfranklin:
Sounds good. :+1:

Issue #35, opened 2015-01-21
Storing element Points in FILE I/O


roigcarlo:
The number of points that can be stored for a given element now is set in the code

Ex: 

~~~~{.python}
MAX_POINTS_IN_FACE = 3
MAX_POINTS_IN_CELL = 4
~~~~

Points are then sotred as

~~~~{.python}
points_uuids = tables.StringCol(32, pos=2, shape=(MAX_POINTS_IN_EDGE,) )
~~~~

We agreed in the last meeting that this need to be changed.

Currently we are considering two options:

* Increase the maximum number to fit the requirements of any element.
* Use VLArray  ( variale lenght arrays ) see  https://github.com/simphony/simphony-common/issues/22#issuecomment-70848123

While I preffer to use VLArray, I can recall that someone argued that this will decrease the performance of the code.

Please let me know what you think.

itziakos:
> While I preffer to use VLArray, I can recall that someone argued that this will decrease the performance of the code.

I also prefer using a VLArray.

The only reason that I can think of is that reading a record will need reading the file in two different points, but on the other hand hdf5 caching and the fact that records will be smaller might negate this behaviour. I would think that performance will depend on the usecase.

nathanfranklin:
VLArray sounds good.  However, I don't totally understand it and am curious how it would be specifically used in this case.

For example, in the table (`self._group.cells`) the ids of the cells are stored.  Would there then be an associated VLArray called something like `self._groups.cell_points` which would contain the point uuids for each cell?  Did I understand that correctly?

If so, then how would the connection between these cells uuids and the rows of the VLArray be made and maintained (e.g. during deletes of cells).  Would the row index of the VLArray (`self._groups.cell_points`) be saved to the row of a particular cell in `self._group.cells`.  

And could the number of points in a particular cell change (e.g. cell1 has 5 points..then later 6 points)?  In the documentation there is "you can only update values with exactly the same size in bytes than the original row." which might be a problem then.

itziakos:
> For example, in the table (self._group.cells) the ids of the cells are stored. Would there then be an associated VLArray called something like self._groups.cell_points which would contain the point uuids for each cell? Did I understand that correctly?

Yes the VLArray will be used to store the list of uuids in the `points` attribute of the `Cell`.

[...]

Issue #20, opened 2014-10-29
Consistent identification (id/uuid) for parameters/attributes in Particle/Point/etc


nathanfranklin:
In Mesh and ParticleContainer hold entities (e.g. Particles, Points, Edges) which have a means of identification.  That identification is used when getting, updating or adding such entities to the high-level objects (Mesh, ParticleContainer).    

This identification should have a consistent name across the classes. Currently, `id` is used as an attribute to Particle/Point/Edge/Cell.  
However in the __init__ methods there are currently some differences:
 *  `id` is used as a parameter in Particle. 
 *   `uuid` is used as a parameter in Point, Edge, Face, and Cell

Example:

from simphony.cuds.mesh import Point
from simphony.cuds.particles import Particle

particle = Particle(**id**=None, coordinates=(0.0, 0.0, 0.0))
point = Point(**uuid**=None, coordinates=(0.0, 0.0, 0.0))

itziakos:
I think it is time to take a vote on the issue. I vote for `uid` based on the https://github.com/simphony/simphony-common/issues/20#issuecomment-67320388

nathanfranklin:
`uid`  :+1:

roigcarlo:
`uid` as well

SGGgarcia:
`uid` sounds good

itziakos:
I am picking this up

itziakos:
work for this milestone done in #61. Moving to next milestone

itziakos:
This has been implemented

PR #36, opened 2015-01-21
Feature mesh io

roigcarlo:
Pull request for the mesh-io feature

An early implementation of the Mesh FILE I/O

**Pending**:
- [x] Store dataContainers for points and elements
- [x] Set the correct number of points for elements (see https://github.com/simphony/simphony-common/issues/35)
- [x] Typos in the comments


simphony/io/cuds_file.py
```diff
@@ -179,3 +261,23 @@ def iter_particle_containers(self, names=None):
...
+
+        Parameters
+        ----------
+        names : list of str
+            names of specific mesh containers to be iterated over.
+            If names is not given, then all meshes will be iterated over.
+
+        """
+
+        names = copy.deepcopy(names)
```
itziakos: no need to deep copy, this line is not necessary since you are not mutating the list of names.

simphony/io/file_mesh.py
```diff
@@ -0,0 +1,736 @@
...
+            Mesh point identified by uuid
+
+        Raises
+        ------
+        Exception
+            If the point identified by uuid was not found
+
+        """
+
+        try:
```
itziakos: The try: expect here will not work since there is no exception raised by the `where` method. I will suggest to use the 
for..else construct in this case see https://github.com/simphony/simphony-common/pull/31/files#diff-48635a0984ef717c0fb53f6c3d74e0fcR112 

also see http://python-notes.curiousefficiency.org/en/latest/python_concepts/break_else.html for a discussion on how it works

simphony/io/tests/test_file_mesh.py
```diff
@@ -0,0 +1,686 @@
...
+
+        points = [
+            self.points[0],
+            self.points[1],
+            self.points[2],
+            self.points[3],
+            self.points[4]
+            ]
+
+        for point in points:
```
itziakos: suggestion (it is not a real problem):  
Using list comprehantions is a common python idiom.
`puuids = [self.mesh.add_point(point) for point in points] `

simphony/io/file_mesh.py
```diff
@@ -0,0 +1,736 @@
...
+            for face in self._group.faces.where('uuid == value',
+                                                condvars={'value':
+                                                          f_uuid.bytes}):
+                return Face(
+                    list(uuid.UUID(bytes=pb) for pb in face['points_uuids']),
+                    uuid.UUID(bytes=face['uuid'])
```
itziakos: Use also `version=4` to designate the this is a type 4 generated uuid

itziakos:
@roigcarlo, I had a first look at the PR, looks good ping me when you wont me to have a look again after you have implemented the pending items

roigcarlo:
@itziakos OK, thank you!

itziakos:
Note that #31 is not available in master and can be used to augment this PR

itziakos:
> Note that #31 is now available in master and can be used to augment this PR

corrected comment

simphony/io/file_mesh.py
```diff
@@ -0,0 +1,763 @@
...
+        Exception
+            If the point identified by uuid was not found
+
+        """
+
+        for row in self._group.points.where(
+                'uuid == value', condvars={'value': p_uuid.hex}):
+            return Point(
+                tuple(row['coordinates']),
+                uuid.UUID(hex=row['uuid'], version=4),
```
roigcarlo: @itziakos, @nathanfranklin I'm having some problems using "bytes" in order to store the representation of the uuid, as you can see in the Travis report (https://travis-ci.org/simphony/simphony-common/builds/49300357). It randomly fails in some cases. Lets say 1 of every 200 tests ends up in a *ValueError: bytes is not a 16-char string* when trying to regenerate the uuid from the field stored

For now, it seems that storing the hex representation instead of the bytes values does the job. Have you any insight of what can be the reason of this behavior?

itziakos:
> For now, it seems that storing the hex representation instead of the bytes values does the job. Have you any insight of what can be the reason of this behavior?

I am not sure why the behaviour is wierd. However, I think that your suggestion makes sense. using the hex representation is more robust

itziakos:
On item that I could think of is that the `bytes` string has a null in the middle which confuses the tables machinery.

itziakos:
I am still not sure about the issue, let me have a more in depth look

itziakos:
@roigcarlo, I checked the offending commit and I see that you define the uuid containing collumns as StringCol(32) instaid of StringCol(16). https://github.com/simphony/simphony-common/commit/dfc1b0ac5100ca0cffd2a55c94128593f49ff151

Could it be this the issue here?

roigcarlo:
I thought the same short after the commit but it happens the same with 16.

I've also preparing some stress test for all classes that make use of tables and uuid right now. I will let you know when the results are ready

itziakos:
@roigcarlo running the DataContainerTable benchmark should be a good stress test. So far, I had not have any problems.

itziakos:
see https://github.com/simphony/simphony-common/blob/master/simphony/bench/data_container_table_bench.py

itziakos:
I would also suggest to enclose the offending code into a try..except block so that you can start the debugger when the error appears again.

Something like
```python
try:
   ....
except ValueError:
    import pdb
    pdb.set_trace()
```

itziakos:
@roigcarlo, I have been running the benchmark code for a while on the DataContainerTable, I have not seen the problem with ``ValueError: bytes is not a 16-char string``. Is it possible that you had some stale code which was using an itemsize of `32` in one of your table description?

itziakos:
If we cannot replicate the error consistently. I would suggest to continue using the `bytes` attribute and open a bug report about the intermittent error and be vigilant next time it appears to gather more info.

It is possible that the underlined error is in the uuid module and not our implementation.

roigcarlo:
@itziakos As you say it appears that the only module affected is this, I believe this is normal, since actually ( correct me if I'm wrong ) I'm the only one that regenerates the uuid from the stored value. Other classes use that value directly, or just don't make use of it at all.

I think the error can be related with that.

After some digging I ended up in this situation. Consider this fragment of execution:

```python
print "Stored:\t \\x" + '\\x'.join(x.encode('hex') for x in cell.uuid.bytes), len(cell.uuid.bytes)
row['uuid'] = cell.uuid.bytes

# some useful code here

try:
    print "Retrieved:\t \\x" + '\\x'.join(x.encode('hex') for x in row['uuid']), len(row['uuid'])
except ValueError:
    import pdb
    pdb.set_trace()
```

A possible output in case the error raises is:

```
Stored:     \xab\xbe\x19\x39\x72\x24\x4e\x68\x92\x44\x21\xc6\x00\x26\xe0\x00 16
Retrieved:  \xab\xbe\x19\x39\x72\x24\x4e\x68\x92\x44\x21\xc6\x00\x26\xe0 15
```

In this case is clear that either pytables or python is just ignoring \x00, provably as is a null character at the end of the string.

But this is not the only time when such a error happens:

```
Stored:      \x2c\xc0\x1e\xd0\x0f\x0f\x4f\x1c\xb1\x68\x22\x69\xae\x5d\x10\x9a 16
Retrieved:   \x2c\xc0\x1e\xd0\x0f\x0f\x4f\x1c\xb1\x68\x22\x69\xae\x5d\x10\x9a 16
```

Here the length is the same and it does not appear to be any problem, but looking at the character value ( notice that i will use <> to delimit a char ):

```
(Pdb) p row['uuid']
<,> <\xc0> <\x1e> <\xd0> <\x0f> <\x0f> <O> <\x1c> <\xb1> <h> <"> <i> <\xae> <]> <\x10> <\x9a>
```

In this case I assume that provably <"> is causing again the string to be miss-parsed by pyTables. I'm not aware of the internal implementation of UUID, but in the doc it suggest that 

```python
UUID(bytes='\x12\x34\x56\x78'*4)
```

Is accepted, so i guess that what is happening is that our value is actually interpreted as an string due to the format in which is stored. and hence pared by python.

So, my suggestion would be, either to use hex, since this seems not to cause any disturbance, or stop using 

```python
tables.StringCol(16, pos=0)
```

Maybe it is interesting to create an Issue, since Lattice and ParticleContainer IO will also face this problem once they implement UUID for FileIO

itziakos:
> @itziakos As you say it appears that the only module affected is this, I believe this is normal, since actually ( correct me if I'm wrong ) I'm the only one that regenerates the uuid from the stored value. Other classes use that value directly, or just don't make use of it at all.

The DataContainerTable is also creating the uuids from the stored value

itziakos:
> So, my suggestion would be, either to use hex, since this seems not to cause any disturbance, or stop using

@roigcarlo, thanks for the analysis. It is really heplfull to understand the issue. I agree with you that using tables.StringCol(32, pos=0)  and using 'hex' instead of 'bytes' will be a good compromise with a little space overhead.

I will open an issue, since it would be good to investigate further for alternatives.

itziakos:
Please ignore comment https://github.com/simphony/simphony-common/pull/36#issuecomment-73230026, it is wrong

roigcarlo:
@itziakos The PR should be ready for review. As you can see I'have also implemented https://github.com/simphony/simphony-common/issues/20 ( and not 22 as it appears in the commit sorry for that )

simphony/io/h5_cuds.py
```diff
@@ -167,3 +244,23 @@ def iter_particle_containers(self, names=None):
...
+
+        Parameters
+        ----------
+        names : list of str
+            names of specific meshes to be iterated over.
+            If names is not given, then all meshes will
+            be iterated over.
+
+        """
+        names = copy.deepcopy(names)
```
itziakos: No need for deepcopy

simphony/io/tests/test_file_mesh.py
```diff
@@ -0,0 +1,603 @@
...
+        """ Check that a point can be retrieved correctly
+
+        """
+
+        puids = [self.mesh.add_point(point) for point in self.points[:1]]
+
+        point_ret = self.mesh.get_point(puids[0])
+
+        self.assertTrue(isinstance(point_ret, Point))
+        self.assertEqual(puids[0], point_ret.uid)
```
itziakos: You should assert that the points have the expected values, in `coordinates` and `data`

simphony/io/tests/test_file_mesh.py
```diff
@@ -0,0 +1,603 @@
...
+        edges = [
+            Edge(puids[:]),
+        ]
+
+        euids = [self.mesh.add_edge(edge) for edge in edges]
+
+        edge_ret = self.mesh.get_edge(euids[0])
+
+        self.assertTrue(isinstance(edge_ret, Edge))
+        self.assertEqual(euids[0], edge_ret.uid)
```
itziakos: see comment https://github.com/simphony/simphony-common/pull/36/files#r24573824

simphony/io/tests/test_file_mesh.py
```diff
@@ -0,0 +1,603 @@
...
+        cells = [
+            Cell(puids[:])
+            ]
+
+        cuids = [self.mesh.add_cell(cell) for cell in cells]
+
+        cell_ret = self.mesh.get_cell(cuids[0])
+
+        self.assertTrue(isinstance(cell_ret, Cell))
+        self.assertEqual(cuids[0], cell_ret.uid)
```
itziakos: similar comment to https://github.com/simphony/simphony-common/pull/36/files#r24573824

itziakos:
Thanks @roigcarlo, the unit-test looks very thorough. There are a few minor comments, apart from that :+1:

simphony/io/h5_cuds.py
```diff
@@ -112,6 +113,45 @@ def add_particle_container(self, particle_container):
         self._file.flush()
         return pc
 
+    def add_mesh(self, name, mesh=None):
```
nathanfranklin: now that the high-level objects have a `name` attribute, this method can be updated to `add_mesh(self, mesh)` similar to `add_particle_container`

nathanfranklin: I guess related to that, one will also need to handle any changes to `name'.  For example,
```
w_mesh_foo = wrapper.get_mesh("foo")
w_mesh_foo.name = "bar"
bar = wrapper.get_mesh("bar") # should work
foo = wrapper.get_mesh("foo") # should fail
```

nathanfranklin: here is the related issue about the `name` attribute:  #38

nathanfranklin:
@roigcarlo  looks good.  just added a few comments related to the `name` attribute

roigcarlo:
All issues should be fixed. @nathanfranklin I would like to address https://github.com/simphony/simphony-common/pull/36#discussion_r24576039 once the wrappers are fully functional if you don't mind

roigcarlo:
@nathanfranklin , I have corrected https://github.com/simphony/simphony-common/pull/36#discussion_r24576039 as we agreed ( and other extra problems that I detected) 
I also added some additional unittest for file_mesh based on the ones for file_particle_container as well

simphony/io/file_mesh.py
```diff
@@ -0,0 +1,778 @@
...
+    data : Data
+        The X coordinate.
+    points : dictionary of Point
+        Points of the mesh.
+    edges : dictionary of Edge
+        Edges of the mesh.
+    faces : dictionary of Face
+        Faces of the mesh.
+    cells : dictionary of Cell
+        Cells of the mesh.
```
nathanfranklin: `name` should be added.  and `points`, `edges`, `faces` and `cells` can be removed.

nathanfranklin:
@roigcarlo  just added one documentation comment.  otherwise looks perfect and good to merge!  :+1:

Issue #24, opened 2014-12-16
Saving and recovering uuids into hdf5


itziakos:
The main issue we are trying to solve here is how to store the uuid 128 bit value into a pytables table

itziakos:
I am reopening the issue, we have seen some intermitten failures with using `uuid.bytes` to store the uuid. The current suggestion is to switch to `uuid.hex` (ie. StringCol(itemsize=32) instead.

itziakos:
Please ignore my previous comment

itziakos:
update in #53

Issue #19, opened 2014-10-15
use uuid for all adapters


itziakos:
Looking at the implementations of the various containers a common pattern is to generate an id using a random number generator and others do not even provide an id. 

Based on the discussion in the simphony wiki high level containers should 
(a) assign an id to items that are added when those items do not have an id assigned and 
(b) use the id of items that have one assigned.

The main problem here is the fact that because (b) there is no control on how the ids are created. So each time case `b` happens an number of ids are randomly generated until one that does not conflict is found (This operation is very expensive for file or remote adapters).

Assume that there are two adapters which assign ids based on internal logic then we cannot be sure that the following will not happen:

```python

mesh1 = adapter1.Mesh()
mesh2 = adapter2.Mesh()

point1 = Point(1, 1, 1)
point1.id = mesh2.add_point(point1)

point2 = Point(0.0, 0.0, 0.0)
point2.id = mesh1.add_point(point2)

# In the generic case where the adapters are not sharing the same process the following
# is not unlikely.
print point2.id != point2.id
```

I would suggest to force all adapters to use the RFC 4122 standard for creating new uuids. This way we can create new uuids without having to worry about identical ones (within practical limits). The RFC 4122 is supported in python (uuid package) and C/C++ so using it will not be a big problem. The only issue I can see is that we have to use 128 bits more for the objects that we instantiate in python, but I think the advantages on speed and code simplicity are going to counter this.

itziakos:
@roigcarlo, is the feature-mesh-io branch ready for review?

itziakos:
moved comment to the right issue.

itziakos:
Currently uuids needs to be supported for 
- [ ] particles
- [ ] mesh
- [ ] particles hdf5 adapter
- [ ] mesh hdf5 adapter
- [ ] lattice*
- [ ] lattice hdf5 adapter*

**Note**:
It is not clear if lattice should use `uuids` or not.

itziakos:
done

PR #37, opened 2015-01-22
[WIP] File lattice

tuopuu:
Pull request for FileLattice class that implements Lattice class in a Cuds file. 

Modified cuds_file.py on the parts related to FileLattice. 

FileLattice now supports variable number of data containers for lattice nodes. Data container definitions are stored in a bitmask implemented as vectors of booleans in an N-dimensional CArray (N=2,3). If more data containers are defined/added later, then the code extends the bitmask by creating a new bitmask with extended vectors, and copies the data from old bitmask to the new one. 

WIP: 
- unit tests
- support for more extensive set of data types for data containers


simphony/io/cuds_file.py
```diff
@@ -85,6 +85,58 @@ def close(self):
...
+            The lattice newly added to the file.  See
+            get_lattice for more information.
+
+        """
+        if name in self._file.root.lattice:
+            raise ValueError(
+                'Lattice \'{n}\` already exists'.format(n=name))
+
+        lat = FileLattice(self, name, lattice._type, lattice._base_vect,
+                            lattice._size, lattice._origin)
```
roigcarlo: This fails the flakes8, there are 2 extra spaces before "lattice"

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,375 @@
+from simphony.core.cuba import CUBA
+from simphony.cuds.lattice import LatticeNode
+
+from tables import *
```
roigcarlo: Please use "import tables"

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,375 @@
...
+
+        Parameters:
+            id : tuple
+                LatticeNode coordinates
+            name : str
+                Name of the PyTables column that is flagged
+        """
+        bits = self._bitmask[id]
+        n = self._ColDescr.columns[name]._v_pos
+        newbits = numpy.array(numpy.zeros(self._mask_size),dtype=bool)
```
roigcarlo: Another flake8 error, a space is required after "(self._mask_size)"

simphony/io/cuds_file.py
```diff
@@ -51,7 +53,6 @@ def open(cls, filename, mode="a", title=''):
         filename : str
             Name of file to be opened.
 
-
```
itziakos: styling nit: Please revert this change, I do not see a reason for having two spaces

simphony/io/cuds_file.py
```diff
@@ -59,7 +60,6 @@ def open(cls, filename, mode="a", title=''):
                 * *'a'*: Append; an existing file is opened for reading and
                   writing, and if the file does not exist it is created.
 
-
```
itziakos: same comments as https://github.com/simphony/simphony-common/pull/37/files#r23379850

simphony/io/cuds_file.py
itziakos: please ignore my comment

simphony/io/cuds_file.py
itziakos: please ignore my comment

simphony/io/cuds_file.py
```diff
@@ -85,6 +85,58 @@ def close(self):
         """
         self._file.close()
 
+    def add_lattice(self, name, lattice, keywords=[]):
```
itziakos: Please set the default argument value of `keywords` to `None`. Using mutable objects for default argument values can lead to very unexpected results in python (see http://docs.python-guide.org/en/latest/writing/gotchas/)

simphony/io/cuds_file.py
```diff
@@ -85,6 +85,58 @@ def close(self):
...
+        """Add lattice to the file.
+
+        Parameters
+        ----------
+        name : str
+            name of lattice
+        lattice : Lattice
+            lattice to be added.
+        keywords : list of str, optional
```
itziakos: I do not see the `keywords` argument used in the code, it does not look that it is necessary.

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,378 @@
...
+
+        # Check that table has all the needed columns
+        if keywords == []:
+            # Add everything
+            self._update_table(node, node.data.keys())
+            for row in self._table.iterrows(n, n+1):
+                if len(node.data) is not 0:
+                    for key in node.data:
+                        name = str(key)[5:]
+                        row[name] = copy.deepcopy(node.data[key])
```
itziakos: There should be no need for deep copy or even copy since the data will be converted/copied into the column data type as necessary.

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,378 @@
...
+            name : str
+                Name of the PyTables column that is flagged
+        """
+        bits = self._bitmask[id]
+        n = self._ColDescr.columns[name]._v_pos
+        newbits = numpy.array(numpy.zeros(self._mask_size), dtype=bool)
+        newbits[n] = True
+        self._bitmask[id] = bits | newbits
+
+    def update_node(self, node, keywords=[]):
```
itziakos: see comment https://github.com/simphony/simphony-common/pull/37/files#r23380264

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,378 @@
...
+                        self._update_bitmask(node.id, name)
+                    row.update()
+        else:
+            # Add only selected keywords
+            self._update_table(node, keywords)
+            for row in self._table.iterrows(n, n+1):
+                if len(node.data) is not 0:
+                    for key in keywords:
+                        name = str(key)[5:]
+                        row[name] = copy.deepcopy(node.data[key])
```
itziakos: see comment https://github.com/simphony/simphony-common/pull/37/files#r23380572

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,378 @@
...
+                        self._update_bitmask(node.id, name)
+                    row.update()
+        else:
+            # Add only selected keywords
+            self._update_table(node, keywords)
+            for row in self._table.iterrows(n, n+1):
+                if len(node.data) is not 0:
+                    for key in keywords:
+                        name = str(key)[5:]
```
itziakos: This looks a bit fragile, there is no guaranty that string name of CUBA keywords will always differ in the first 5 characters.

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,378 @@
...
+                        row[name] = copy.deepcopy(node.data[key])
+                        self._update_bitmask(node.id, name)
+                    row.update()
```
itziakos: you also need to `flush` the table when the update is finished.

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,378 @@
...
+                if len(node.data) is not 0:
+                    for key in node.data:
+                        name = str(key)[5:]
+                        row[name] = copy.deepcopy(node.data[key])
+                        self._update_bitmask(node.id, name)
+                    row.update()
+        else:
+            # Add only selected keywords
+            self._update_table(node, keywords)
+            for row in self._table.iterrows(n, n+1):
```
itziakos: `row = self._table[n]` should be enough.

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,378 @@
...
+            self._origin = self._table.attrs.origin
+            # Update column description class.
+            data = {}
+            for key in self._table.coldescrs:
+                if not key == 'INDEX':
+                    data[eval('CUBA.'+key)] = self._table[:][key][0]
+            if not data == {}:
+                self._update_coldescr(data, data.keys())
+
+    def _construct_default_rowclass(self):
```
itziakos: I would suggest to use the `DataContainerTable` implementation in #31 (soon to be merged) to store the `data` attribute of the nodes.

itziakos:
@tuopuu, I have a couple of generic comments

- Ids should be uuid.uuid4() #19 and #24
- I think you can greatly simplify the implementation by re-using the DataContainerTable implementation in #31. It will for sure need lest testing since DataContainerTable is already tested.

itziakos:
Please note that #31 is now available in master

itziakos:
> I think Ids should be uuid.uuid4() #19 and #24

hmm.. looking at the code again it is a little confusing that the node.id is supposed to be the x,y,z coordinates. This makes the Lattice container quite different than the rest @nathanfranklin, does #19 and #24 apply to the lattice?

simphony/io/file_lattice.py
itziakos: The DataContainerTable behaves like a dictionary 
```python
table = DataContainerTable(root, 'node_data')
data = DataContainer()
# add a new data container to the hdf5 table
uid = table.append(data)
# get a container back
data = table[uid]
# delete a data container
del table[uid]
```
So using the DataContainerTable you only need to save a mapping between the node coordinates and the `uid` returned by append.

kemattil:
In the lattice data specification it was agreed that the LatticeNode id is given by the node index coordinate (unique for a given lattice). But of course we can define a mapping between an uid and a node index coordinate when storing data to pytables (in fact, I think Tuomas already uses some kind of mapping between row numbers and node index coordinates); node index coordinates would then be stored not in the id column but instead in a separate coordinate column.

nathanfranklin:
> node.id is supposed to be the x,y,z coordinates 

As @kemattil  mentioned, the node.id is node-index-coordinates.  For a user to get the x,y,z coordinates of the node, he/she has to use the `get_coordinate` (or alternatively calculate the x,y,z coordinates by using the base vectors).

>> I think Ids should be uuid.uuid4()  simphony/simphony-common#19 and simphony/simphony-common#24

> hmm.. looking at the code again it is a little confusing that the node.id is supposed to be the x,y,z coordinates. This makes the Lattice container quite different than the rest @nathanfranklin, does #19 and #24 apply to the lattice?
 

My opinion is that `uid` is not needed for these lattices as these are in a fixed-grid and having `uid` would add an additional (in my opinion, unneeded)  naming identification when the node-index is already sufficient.  For example: 
```python
l = lattice.make_square_lattice('lattice', 0.2, (3, 2))

for node in l.iter_nodes():
    print node.id
```
Here it is clear that there is 3x2 lattice.  So there are 6 nodes with the following node indices :   (0, 0) (0, 1) (1, 0) (1, 1) (2, 0) (2, 1).  These node indices provide a unique identification of all nodes in the lattice.

So I feel simphony/simphony-common#19 and simphony/simphony-common#24 don't apply to lattice.

-------------------------------

We could change the `node.id` to  `node.index` or `node.index_coordinate`

tuopuu:
We've been developing Lattice / FileLattice on the assumption that once initiated, the lattice dimensions will not be changed. (In a typical simulation setup, this never happens so it's a sensible assumption.) Therefore, ordering of data in a PyTables table will be fixed, and so we can define a simple and efficient ND <-> 1D bijection. This mapping is used for example between the bitmask (that flags defined CUBA keys for each LatticeNode in the current implementation) and PyTables table that contains CUBA key data on its columns. 

I realize that having the same convention of id'ing elements in each of the discrete systems (mesh, lattice, particles) would be elegant on the SimPhoNy perspective, but on the other hand, (i,j,k) indexing of lattice nodes is a very well agreed convention in lattice-Boltzmann community. My opinion is that we keep (i,j,k) indexing, although I'm not against changes now or in future. :)

As @nathanfranklin suggested we really could further dissociate node.id from uid by renaming it e.g. to node.index

itziakos:
@nathanfranklin, @tuopuu and @kemattil, Thanks for the explanation, from what I understand there is  no objection with id not been uuid although I would prefer `index` from `id` for this case.

kemattil:
I agree, node.index is a better name.

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,378 @@
...
+        PyTables table for the lattice can be generated. Empty table
+        cannot be created.
+        TODO: Find a more compact (and still meaningful) way of
+        generating a dense table.
+        """
+        self._ColDescr = type("IsDescription", (tables.IsDescription,),
+                              {'INDEX': tables.Int32Col(pos=0,
+                               shape=len(self._base_vect))})
+
+    def _update_coldescr(self, data, keywords):
```
itziakos: One can retrieve the full `DataContainer` table description from https://github.com/simphony/simphony-common/blob/master/simphony/io/data_container_description.py#L5

tuopuu:
@itziakos Do you suggest using the full 66 column table always in a Cuds file when storing data, even though we only might use 2 CUBA keys for lattice-Boltzmann output (density & velocity)? Do you know if there is performance penalty between these two cases, 2 Columns vs. 66 columns with zeroes? 

If the CUBA datatypes are predefined, how could we save the lattice-Boltzmann distributions (n-dimensional float64 vector, where n=19 for example)? The need for saving the distribution values to a file comes useful if LB-solver is needed to be restarted. Should we define just new CUBA keys for all of these LB-models, even that it further expands the data table for all other users?

simphony/io/file_lattice.py
itziakos:
```python
from simphony-common.io.data_container_description import Record
DataDescription = Record.columns['Data']
```

itziakos:
> @itziakos Do you suggest using the full 66 column table always in a Cuds file when storing data, even though we only might use 2 CUBA keys for lattice-Boltzmann output (density & velocity)? Do you know if there is performance penalty between these two cases, 2 Columns vs. 66 columns with zeroes?

In theory there might be a small performance hit for reading larger columns but the size of the cache, the access mode (using an index for example) and other parameters take effect so unless we actively test the different configuration we cannot compare. 

> If the CUBA datatypes are predefined, how could we save the lattice-Boltzmann distributions (n-dimensional float64 vector, where n=19 for example)? The need for saving the distribution values to a file comes useful if LB-solver is needed to be restarted. Should we define just new CUBA keys for all of these LB-models, even that it further expands the data table for all other users?

I do not know the particulars of this issue (maybe @nathanfranklin could provide more). Please open a separate github issue and note there is a lot of discussion regarding CUBA keys, DataContainers and storage policies (see https://github.com/simphony/simphony-lammps-md/pull/5#discussion_r23304088, #29 and #35). I personally prefer to have a better description on what is used/supported. It is currently a hot topic and will need to be resolved soon. 

Regarding this PR:
Using the current implementation of the DataContainerTable will simplify the implementation of the Lattice file io in the following ways:
- No need to create the table description dynamically
- Removes a good amount of code that it currently a little fragile and untested and will take some time to cleanup.

Please note that the DataContainerTable is built to be a generic object and can be adapted to support saving only a fraction of the full CUBA keywords. Please let me know if such a feature will be useful to your implementation.

kemattil:
I agree that using a common DataContainerTable is a good idea (for the reasons @itziakos mentioned). However, it is extremely difficult for me to approve storing values for, e.g., 66 keywords when we actually need to store only for 4 or so - with or without compression. Hence, we need a feature which allows saving only a fraction of the keywords (e.g. an argument keywords=None for the DataContainerTable constructor).

The number of components in a distribution vector is a separate issue which we need to resolve somehow. That is, the number of components depends not only on the spatial dimension, but also on the specific lattice-Boltzmann model utilized: e.g. typical models in 2D use 7,9,17,25,and 37 components while in 3D the usual models involve 13,15,19,27,39, etc. components.

tuopuu:
If the table is stored and compressed column-wise, in large chunks, the ~66 zero columns should not become a problem from an increased file size point of view. At best, the zeros could be expressed with ~66 ((long) long) integers, which is irrelevant compared to disk space requirements for filled columns. My only concern is that compression algorithms available in PyTables/HDF5 may not be optimal for storing such large sets of zeros because of chunking. On the other hand, zero columns could also cause performance issues, but that should be tested at some point.

However, if we decide not to store the whole set of ~66 columns of zeros, then we have decide between (at least) two choices: 1) ban adding of columns to an existing table, which then forces recreation of the table and copying data from the old one to the new one when new columns are needed, or 2) implement dynamic extension of a table, which is something we already have drafted in the current version of FileLattice.

Option 1 sounds weird to me, because in lattice-Boltzmann simulations only the geometry could take tens or hundreds of GB's of disk space (using non-optimal compression algorithms), and unnecessary copying of such dataset will be a bottle neck in input/output. Perhaps we should still consider the option of storing CUBA key-data to separate tables instead of columns in a single table?

kemattil:
I'm not in favor of storing all CUBA keywords. Note also that a single keyword does not represent a scalar in general: it can represent a vector (with and arbitrary number of components) or perhaps even a tensor in the future. Also, we can expect the CUBA keyword-list to extend in the future (this is an issue even if the keywords are assigned into subgroups in some manner).

I still think that a DataContainerTable with a specified list of CUBA keywords is a good compromise solution. In a typical scenario, the keywords utilized are known/specified a priori (or I couldn't immediately come up with a situation where the variables utilized are not known beforehand). Hence, the demand for dynamically extending a table with new keywords should not arise in a basic scenario.

Storing CUBA key-data to separate tables, instead of columns in a single table, would probably better reflect the chosen (lattice) data structures, but, as we have already discussed, is not trivial to implement. Hence, at this point, in order to progress, I prefer the DataContainerTable with a specified list of CUBA keywords.

tuopuu:
After examining DataContainerTable I also think that it is a good compromise to adopt it as the basis of in-file data structure for lattice-systems, as @itziakos suggested. However, I can't really start the adoption before the extensions suggested in #44 are implemented. It seems useless to continue with the current version of dynamic table, so @itziakos, please feel free to assign some of the work in #44 to me, if you don't have time for it yourself.

kemattil:
Excellent, we have reached a consensus! It would also be great, in many ways, if @tuopuu you could review the pull request when the extensions are ready, i.e. in the case @itziakos implements them.

tuopuu:
We can now close this PR as there is a new branch for FileLattice #67.

PR #43, opened 2015-01-23
add requirements to the setup.py

itziakos:


itziakos:
@dpinte can I get a quick review

setup.py
```diff
@@ -30,6 +30,12 @@ def write_version_py(filename=None):
     author='SimPhoNy FP7 European Project',
     description='The native implementation of the SimPhoNy cuds objects',
     long_description=README_TEXT,
+    install_requires=[
+        "cython",
```
dpinte: is cython in install_requires or setup_requires?

dpinte:
@itziakos except one comment, this LGTM

setup.py
itziakos: cood point. cython is actually only necessary for th depedencies not for building simphony-common it-self.

Issue #39, opened 2015-01-22
Add an abstract interface for the Lattice container


itziakos:

itziakos:
@tuopuu we need an abstract definition of the class similar to the mesh and particile containers

kemattil:
This was left open by intention (originally it was not considered necessary). However, during our last teleconference with Adham & Nathan, we decided to implement it. It is still in the list of things to do. :)

itziakos:
Please note that this is a high priority item for me since an abstract class with appropriate documentation will provide me with the template for verifying conformance the api of the different engine implementations.

kemattil:
@tuopuu , do you have time this week? Alternatively, I will do it at the beginning of next week (at the latest). The variable node.id should be renamed to node.index in this update.

tuopuu:
Yes, I have time for this, but I have been thinking would it be easier just to add "file=None" optional, last, parameter to abstract Lattice class constructor, so that it would also specify the FileLattice class? To my understanding FileLattice should be an implementation of Lattice class in a Cuds-file. All the other methods in the abstract Lattice are the same in FileLattice. What's your opinion on this, @kemattil and @itziakos?

kemattil:
In my opinion, the "file=None" should be a parameter for the FileLattice class constructor only, i.e. not for the ABCLattice constructor. The constructor of a derived/inherited class can have more parameters than the base class.

itziakos:
implemented in #45

PR #45, opened 2015-01-26
Abclattice

tuopuu:
Added ABCLattice abstract class and renamed id to index as was discussed in #39.


itziakos:
I had a quick look:
- Please add the docstrings to the ABCLattice
- The Lattice class should either:
 -  subclass from `ABCLattice` or 
 -  subclass `object` and register itself as implementing the ABCLattice interface.

itziakos:
I see that you fixed the second case

simphony/cuds/abstractlattice.py
```diff
@@ -0,0 +1,77 @@
...
+        ABCLattice ---> Common Base abstract class ("interface") for
+            the Lattice.
+"""
+
+from __future__ import print_function
+from abc import ABCMeta, abstractmethod
+
+
+class ABCLattice(object):
+    """Abstract base class for a Lattice item."""
```
nathanfranklin: Should the `name` attribute be added to the docstring?

How about the other attributes used by Lattice ( `type, base_vect,  size, origin`)?  I assume they will be supported by FileLattice and any other classes implementing the ABCLattice interface.

itziakos: Yes, I agree, the name is not part of all top level cuds containers

itziakos: updated comment ``not -> now``

nathanfranklin:
@tuopuu @kemattil   This looks good.  Can you address the remaining comments so that this can be merged?

tuopuu:
I replaced abstractlattice.py with the version from @nathanfranklin 's abclattice_update branch, and added type, size, origin and base_vect attributes to docstrings (as @nathanfranklin suggested). These attributes are essential in defining a lattice and should be therefore always associated with the representing object.

nathanfranklin:
@tuopuu  Thanks.  I will fix merge conflict in lattice.py and then merge

tuopuu:
@nathanfranklin OK. What was causing a conflict if I may ask? Please note that there are some changes to lattice.py in this branch that should also be merged to master.

nathanfranklin:
@tuopuu  there was a conflict with simphony/cuds/lattice.py when merging.  please review the commits then we we can merge this pr and close it.

tuopuu:
lattice.py is fine now: I just changed size parameter to tuple as it is in ABCLattice description. In my opinion this merge is good to go.

PR #49, opened 2015-02-02
fix install_requires -> requires

itziakos:
This PR fixes the setup which was using install_requires by mistake


itziakos:
@dpinte can you have a look?

dpinte:
@itziakos the change looks trivial but I can't find any documentation argumenting in favor of using `requires` in place of `install_requires`.  Do you have anything better than this? https://pythonhosted.org/setuptools/setuptools.html#declaring-dependencies Or is this just for depending only on distutils and not setuptools?

itziakos:
@dpinte, I think you are right. This PR is wrong, thanks

Issue #44, opened 2015-01-26
Extensions to DataContainerTable


itziakos:
The following extension to the DataContainerTable are necessary

- [ ] Support version 
- [x] Control which keywords are necessary to be saved/restored

itziakos:
removing `- [ ] Support a factory function to be used as a key.` since currently there is no need for this feature.

PR #50, opened 2015-02-02
Custom record for data container table

itziakos:
This PR adds support for defining a custom record description to be used for the DataContainerTable (see #44).

in more detail:
- The DataContainerTable now accepts an optional argument that defines a custom record to be used for the table
- Two new testcases that check functionality have been added. The testcases are based on abc mixin class has been factored out from the original DataContainerTable testcase.
- Tests have been agumented


itziakos:
@tuopuu, This PR is ready for review

tuopuu:
Looks good to me. uid-indexing looks good for FileMesh/FileParticleContainer, but it's not going to be used for FileLattice. That's not a problem, though, because we can use _retrieve/_populate to read/write data to a specific row number corresponding to the fixed indexing of the lattice nodes.

itziakos:
> That's not a problem, though, because we can use _retrieve/_populate to read/write data to a specific row number corresponding to the fixed indexing of the lattice nodes.

I would advise against using the private methods _retrieve and _populate and instead use the uids internally. Please note that one can construct the UUID instance manually and give it any value 

```python
uuid.UUID(int=1)

```

Please also note that using uuids will be closer to the sharing feature that is currently been discussed.

tuopuu:
Because the set of lattice nodes can be ordered (unlike particles in some cases), we can use this ordering to store the data. We don't have to search for a specific uid from the list of uids ("index" column) to find the corresponding row, when data from that row is requested. Instead, we can immediately read the contents of the row given by this ordering, without explicit searches for uids done in DataContainerTable.__getitem__. However, we can initialize uid in a logical way, e.g. by giving the rows their natural ordering number as a uid, so that for example DataContainerTable.itersequence can be used if really needed for some reason.  

As a compromise, could you consider adding public versions of _retrieve/_populate methods so that access to specific rows would be done legally?

itziakos:
> As a compromise, could you consider adding public versions of _retrieve/_populate methods so that access to specific rows would be done legally?

I will have PR with this feature shortly

PR #48, opened 2015-01-30
add 'name' attribute to high-level CUDS objects

nathanfranklin:
This addresses #38.  A read/write `name` attribute is added to high-level CUDS objects.  This `name` is now used by CudsFile (get_particle_container...etc).

- [x] add 'name' attribute to Mesh
- [x] add 'name' attribute to ParticleContainer 
- [x] add 'name' attribute to FileParticleContainer
- [x] use 'name' of particle container in CudsFile

Issue #47, opened 2015-01-29
add ABCModelingEngine to simphony/cuds


nathanfranklin:
https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/Wrapper_description#Interface_for_wrapper_to_a_modeling_engine

itziakos:
I think that the name should be `ABCModelingEngine`

kemattil:
I agree, ABCModelingEngine is a better name.

Issue #38, opened 2015-01-22
 add 'name' attribute to high-level cuds objects


nathanfranklin:
Each high-level cuds object (ParticleContainer, Mesh, Lattice) would have a `name`  attribute

nathanfranklin:
this issue replaces and closes #30

nathanfranklin:
with PR #48, `name` attribute has been added to Mesh, ParticleContainer and FileParticleContainer

nathanfranklin:
with #36 ,  all high-level objects in simphony-common have teh name attribute.

PR #52, opened 2015-02-03
add ABCModelingEngine

nathanfranklin:
This addresses #47.

- [x] create AbcModelingEngine based on description in wiki
- [x] update use of `name` paramter in methods based on  #48 and #38
- [x] add docstrings

### DIFF
--- a/simphony/cuds/abc_modeling_engine.py
+++ b/simphony/cuds/abc_modeling_engine.py
@@ -1,0 +1,69 @@
+from abc import ABCMeta, abstractmethod
+
+
+class ABCModelingEngine(object):
+    """Abstract base class for a modeling engines in SimPhoNy
+
+    Attributes
+    ----------
+    BC : DataContainer
+        container of attributes related to the bondary conditions
+    CM : DataContainer
+        container of attributes related to the computational method
+    SP : DataContainer
+        container of attributes related to the system paremters/conditions
+
+    """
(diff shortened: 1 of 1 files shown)


itziakos:
@nathanfranklin, ping me when this PR is ready for review.

nathanfranklin:
@itziakos ,  this is ready for PR.

I have two concerns:
 - location: is simphony-common/cuds/abc_modeling_engine.py correct?
 - docstrings: how does the doc string sound for the get_XXX (and add_XXX) methods sound

itziakos:
> location: is simphony-common/cuds/abc_modeling_engine.py correct?

I think that this is ok

simphony/cuds/abc_modeling_engine.py
```diff
@@ -0,0 +1,213 @@
...
+            given, then all meshes will be iterated over.
+
+        """
+        pass
+
+    @abstractmethod
+    def iter_particle_containers(self, names=None):
+        """ Returns an iterator over a subset or all of the particle containers.
+
+        See get_particle_container for more information.
```
itziakos: I do not think that his give more info, I would suggest that we remove such comments

itziakos:
minor comments apart from that :+1:

nathanfranklin:
Thanks for the quick response.  Just addressed the comment.

PR #54, opened 2015-02-04
Row access data container table

itziakos:
This PR provides an additional pytable table proxy for DataContainers. The IndexDataContinersTable supports the Sequence api and it is optimized for faster read and update operations relative to the normal DataContinerTable which is index using uuid. The limitation is that the delete operations cannot be performed.

@tuopuu, I think should suite better the lattice usecase. 

Please note that there is plenty of space for optimizations but, I would suggest that these take place at a later time and using more realistic usecases.


itziakos:
@tuopuu, @nathanfranklin  and @roigcarlo, I have updated the PR so that the table description is automatically created by the cuba-generate script

itziakos:
@tuopuu, is this PR ok to merge?

tuopuu:
Yes, it's OK to merge. Current features should be enough to implement FileLattice.

itziakos:
Thanks @tuopuu

PR #56, opened 2015-02-09
Doc api reference

itziakos:
This PR add automatic extraction of the docstrings of the main cuds submodules into the sphinx documentation

Pending items
- [x] Update docstrings to conform with the numpydoc standard (see https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard)


itziakos:
@nathanfranklin  and @roigcarlo. This PR is now ready for review:
- I have added a new command in setup.py that will build the documentation (assuming that the new dev_requirements are satisfied).
- The api reference section is generated directly from the code docstrings
- I have also did some minor cleanup of the docstrings to follow a commonly used standard (https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard)

simphony/cuds/abc_modeling_engine.py
```diff
@@ -37,16 +37,9 @@ def add_lattice(self, lattice):
 
         Parameters
         ----------
-        lattice: ABCLattice
+        lattice : ABCLattice
             lattice to be added.
 
-        Returns
```
nathanfranklin: Is the "Returns" section removed from the add_XXX methods because one can already read it in get_XXX()?

doc/source/api/cuds.rst
```diff
@@ -0,0 +1,67 @@
...
+.. rubric:: Classes
+
+.. currentmodule:: simphony.cuds
+
+.. autosummary::
+
+   ~lattice.LatticeNode
+   ~lattice.Lattice
+   ~particles.Particle
+   ~particles.ParticleContainer
```
nathanfranklin: ~particles.Bond

doc/source/api/io.rst
```diff
@@ -0,0 +1,50 @@
+File-IO
```
nathanfranklin: Could we change this to "Cuds-File", or something else.  This might be a dumb nitpick that can be ignored but was just wondering if we could avoid saying the term FILE-IO to avoid confusion with 'FILE-IO wrapper'

simphony/cuds/abc_modeling_engine.py
itziakos: If I am correct, adding a lattice into a lattice container should not return anything.

doc/source/api/io.rst
itziakos: I would prefer to change the name of CUDS File into H5CUDS for the following reasons:
- The name is specific and help avoiding confusion with the FILE-IO based engine adapters
- The name allows alternative serialization types to be provided in the future (e.g SQLCUDS)

Then the section can be named accordingly

doc/source/api/cuds.rst
itziakos: Fixed

simphony/cuds/abc_modeling_engine.py
nathanfranklin: add_lattice, add_particle_container and add_mesh in AbcModelingEngine would be like add_lattice, add_particle_container and add_mesh methods in CudsFile.  The methods would add a mesh, lattice or pc to the engine and the method would return a mesh, lattice, or pc which can then be used to update/query the internal representation. 


Comes from https://github.com/simphony/simphony-common/pull/5#issuecomment-55582063:
> 2. return a handle when adding container file

doc/source/api/io.rst
nathanfranklin: Sounds good.

simphony/cuds/abc_modeling_engine.py
itziakos: I thought that this comment was specific to the CudsFile object and not to the generic engine 
adapter.

- According to the wiki discussion when I add a CUDS container into an engine only the compatible data are transferred to the internal engine. That means that when the a data container with the `name` is requested from an engine it is not guarantied that it will contain all the original information. Having the add_XXX methods return a proxy object could confuse since in-general you would not expect that an add method would return a similar object back.

```python
mesh = Mesh('my_mesh')
engine_mesh = engine.add_mesh(mesh)
```
instead of the more specific

```python
mesh = Mesh('my_mesh')
engine.add_mesh(mesh)
engine_mesh = engine.get_mesh('My mesh')
```

However, I can see that this should probably go through some discussion. So I will revert the changes and add a question issue

nathanfranklin: >Having the add_XXX methods return a proxy object could confuse since in-general you would not expect that an add method would return a similar object back.
yeah, i can see that could be confusing.  yeah, issue to discuss this sounds good!

itziakos:
@nathanfranklin, this PR is ready for another round of reviews. the automaticl build now works as expected http://simphony.readthedocs.org/en/doc-api-reference/

nathanfranklin:
Looks good! :+1:

itziakos:
thanks @nathanfranklin

PR #57, opened 2015-02-10
fix flake8 error in master

itziakos:


itziakos:
@nathanfranklin or @roigcarlo, the is a new version of flake8 which checks on more things and thus the master was faling. This PR fixes master

nathanfranklin:
:+1:

Issue #55, opened 2015-02-06
Cannot reliably store uuids using the `bytes` attribute


itziakos:
@roigcarlo, has seen intermittent problems with using the bytes attribute for recovering the uuid stored as a 16 character byte string. Please find below the analysis of the issue.

After some digging I ended up in this situation. Consider this fragment of execution:

```python
print "Stored:\t \\x" + '\\x'.join(x.encode('hex') for x in cell.uuid.bytes), len(cell.uuid.bytes)
row['uuid'] = cell.uuid.bytes

# some useful code here

try:
    print "Retrieved:\t \\x" + '\\x'.join(x.encode('hex') for x in row['uuid']), len(row['uuid'])
except ValueError:
    import pdb
    pdb.set_trace()
```

A possible output in case the error raises is:

```
Stored:     \xab\xbe\x19\x39\x72\x24\x4e\x68\x92\x44\x21\xc6\x00\x26\xe0\x00 16
Retrieved:  \xab\xbe\x19\x39\x72\x24\x4e\x68\x92\x44\x21\xc6\x00\x26\xe0 15
```

In this case is clear that either pytables or python is just ignoring \x00, probably as is a null character at the end of the string.

But this is not the only time when such a error happens:

```
Stored:      \x2c\xc0\x1e\xd0\x0f\x0f\x4f\x1c\xb1\x68\x22\x69\xae\x5d\x10\x9a 16
Retrieved:   \x2c\xc0\x1e\xd0\x0f\x0f\x4f\x1c\xb1\x68\x22\x69\xae\x5d\x10\x9a 16
```

Here the length is the same and it does not appear to be any problem, but looking at the character value ( notice that i will use <> to delimit a char ):

```
(Pdb) p row['uuid']
<,> <\xc0> <\x1e> <\xd0> <\x0f> <\x0f> <O> <\x1c> <\xb1> <h> <"> <i> <\xae> <]> <\x10> <\x9a>
```

In this case I assume that probably <"> is causing again the string to be miss-parsed by pyTables. I'm not aware of the internal implementation of UUID, but in the doc it suggest that 

```python
UUID(bytes='\x12\x34\x56\x78'*4)
```

Is accepted, so i guess that what is happening is that our value is actually interpreted as an string due to the format in which is stored. and hence pared by python.

So, my suggestion would be, either to use hex, since this seems not to cause any disturbance, or stop using 

```python
tables.StringCol(16, pos=0)
```

[...]

PR #58, opened 2015-02-11
Use uuid hex instead of bytes in data table

itziakos:
This PR updates the uuid based data container Table to use the `hex` representation of the uuid (see #55) for an explenation of the problems with `bytes`

### DIFF
--- a/simphony/core/cuba.yml
+++ b/simphony/core/cuba.yml
@@ -1,9 +1,9 @@
-- description: Universal unique id
+- description: Universal unique id represented as a hex string size 32
   domain: [ATM, DEM, FEM, FVM, LBM, SPH, VIS]
   key: ID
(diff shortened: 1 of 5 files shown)


itziakos:
@roigcarlo, I have updated the uuid based data container table to use hex, can you give a quick review?

roigcarlo:
Seems good but, I think we should update the definition on the yml here as well: https://github.com/simphony/simphony-common/blob/master/simphony/core/cuba.yml#L6

itziakos:
Good point, fix is on the way.

itziakos:
@roigcarlo, PR updated please have another look

roigcarlo:
:+1:

itziakos:
Thanks

PR #67, opened 2015-02-18
PR: File lattice2

tuopuu:
FileLattice now supports CustomRecord.


tuopuu:
@roigcarlo, @nathanfranklin, @itziakos this PR is ready for review. Although, I don't know how to fix Travis CI build...

itziakos:
@tuopuu there are two problems 

flake8 reports these style problems
```
$ flake8 .
./simphony/cuds/particles.py:463:21: W503 line break before binary operator
./simphony/cuds/particles.py:474:28: W503 line break before binary operator
./simphony/cuds/particles.py:482:28: W503 line break before binary operator
```

And building the sphinx documentation reports two problems with the docstrings section styling

```
/home/travis/build/simphony/simphony-common/simphony/cuds/lattice.py:docstring of simphony.cuds.lattice.Lattice:6: SEVERE: Unexpected section title.
Attributes:
-----------
/home/travis/build/simphony/simphony-common/simphony/cuds/lattice.py:docstring of simphony.cuds.lattice.MetaLattice:7: SEVERE: Unexpected section title.
Attributes:
-----------
```

The error is that you have added a `:` at the end of the Attributes subsection this is not following the https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt style that the documentation renderer knows how to parse.


The flake8 error is the one that makes the PR fail, the docstring style error is not making the build fail but it would be good to fix so that the html documentation can be properly compiled.

simphony/cuds/lattice.py
```diff
@@ -25,30 +24,33 @@ def __init__(self, index, data=None):
...
+    DataContainers.
+
+    Attributes:
+    -----------
+    name: string
+        name of the lattice
+    type: string
         Bravais lattice type (should agree with the base_vect below).
-    base_vect : D x float
+    base_vect: D x float
```
itziakos: please leave the space before `:` it is of the numpydoc string guideline

itziakos:
@tuopuu 

> flake8 reports these style problems 

Correction it looks that the error is in current master. I will open a bug report for that.

simphony/cuds/lattice.py
```diff
@@ -25,30 +24,33 @@ def __init__(self, index, data=None):
             self.data = DataContainer(data)
 
 
-class Lattice(ABCLattice):
-    """
-    A Bravais lattice;
-    stores references to data containers (node related data).
+class MetaLattice(ABCLattice):
```
itziakos: **Naming**: I would avoid of the `Meta` in class names unless they are actually defining a python meta-class. I think probably a `BaseLattice` is a better name for the common ancestor of Lattice and FileLattice

simphony/cuds/lattice.py
```diff
@@ -67,6 +69,43 @@ def origin(self):
         return self._origin
 
     def get_node(self, index):
+        pass
```
itziakos: **Querstion**: These methods are already defined in ABCLattice, redefining them here into empty overrides the requirements for them to be defined and does not provide any additional functionality. It is not clear why this is necessary.

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,194 @@
...
+
+        Returns:
+        -----------
+        A reference to a LatticeNode object
+
+        """
+        n = self._get_row_number(index)
+        dcs = {}
+        colnum = 0
+        for column in xrange(self._table._table.description.mask.shape[0]):
```
itziakos: self._table[n] returns a DataContainer with all the information set. I do not see why it is necessary to perform this block of code. Am I missing something?

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,194 @@
...
+            if self._table._table[n][0][column]:
+                data = self._table._table.description.data
+                key = eval(CUBA.__name__+'.'+data._v_names[colnum].upper())
+                dcs[key] = self._table[n][key]
+            colnum += 1
+        return LatticeNode(index, dcs)
+
+    def update_node(self, node, keywords=None):
```
itziakos: This is not following the right signature there is no `keywords` argument in `update_nodes`

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,194 @@
...
+                reference to LatticeNode object
+            keywords : list, optional
+                List of CUBA-keywords that are updated. If empty list
+                then all DC's are from the node
+
+        """
+        # Find correct row for node
+        n = self._get_row_number(node.index)
+        colnms = self._table._table.description.data._v_names
+        if keywords is None:
```
itziakos: Why not do:
```python
updated = {key: value for key, value in node.data.iteritems() if key in keyword}
data = self._table[n] 
data.update(updated)
self._table[n] = data
```

However please note that `keywords` is not part of the `update_node` signature. The following code should be probably more appropriate
```python
self._table[n] = node.data
```

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,194 @@
...
+        indices: iterable set of D x int, optional
+            node index coordinates
+
+        Returns:
+        -----------
+        A generator for LatticeNode objects
+
+        """
+        if indices is None:
+            for row in np.ndenumerate(self._table):
```
itziakos: why not:
```python
for row_number, data in enumerate(self._table):
    index = self._get_index(row_number)
    yield LatticeNode(index, data)

File: simphony/io/h5_cuds.py [hunk 8]
```diff
@@ -152,6 +155,47 @@ def add_mesh(self, mesh):
         self._file.flush()
         return m
 
+    def add_lattice(self, lattice, record=None):
```
itziakos: If I am correct `record` is not part of the interface

simphony/io/h5_cuds.py
```diff
@@ -194,6 +238,28 @@ def get_mesh(self, name):
...
+        no longer be used.
+
+        Parameters
+        ----------
+        name : str
+            name of lattice to return
+        """
+        if name in self._file.root.lattice:
+            F = self._file.root.lattice._v_leaves[name].attrs
+            lat = FileLattice(self, name, F.type, F.base_vect,
```
itziakos: why pass `self` here?

itziakos:
@tuopuu, I have a first look and made some comments, but in general I would be also good to have a generic summary of what one expect in a PR  and a line or two about the changes. 

From my review It looks that the there are a number of changes related to the current CUDS description. It is not clear why they were necessary, maybe @nathanfranklin can understand it better. I am also a little confused about the use of IndexDataContainerTable, I made some suggestions but I might be missing something.

simphony/io/h5_cuds.py
```diff
@@ -152,6 +155,47 @@ def add_mesh(self, mesh):
...
+        ----------
+        FileLattice
+            The lattice newly added to the file.  See
+            get_lattice for more information.
+
+        """
+        if lattice.name in self._file.root.lattice:
+            raise ValueError(
+                'Lattice \'{n}\` already exists'.format(n=lattice.name))
+        filelat = FileLattice(self, lattice.name, lattice.type,
```
itziakos: Why pass self here?

simphony/cuds/lattice.py
itziakos: update link to the numpy docstring guideline https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

simphony/io/h5_cuds.py
```diff
@@ -152,6 +155,47 @@ def add_mesh(self, mesh):
...
+                              lattice.origin, record)
+
+        # copy the contents of the lattice to the file
+        if isinstance(lattice, Lattice):
+            if record is None:
+                for node in lattice.iter_nodes():
+                    filelat.update_node(node)
+            else:
+                keywords = record.columns['data']._v_names
+                ckeys = list({eval(CUBA.__name__+'.'+k.upper())
```
itziakos: using `eval` to get attributes is a little crud (also eval is use very really in production code). It also not necessary to use a set and then create a list.
```python
[getattr(CUBA.__name__, k.upper()) for k in keywords]
```

itziakos: Please also note that the CUBA enum has as members argument that can be used to make mapping between names and keys. See examples:
- https://github.com/simphony/simphony-common/blob/master/simphony/io/indexed_data_container_table.py#L58
- https://github.com/simphony/simphony-common/blob/master/simphony/core/data_container.py

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,194 @@
...
+        index: D x int (node index coordinate)
+
+        Returns:
+        -----------
+        D x float
+
+        """
+        return self._origin + self._base_vect*np.array(index)
+
+    def _get_index(self, n):
```
itziakos: I think that these operations can be performed by numpy 
http://docs.scipy.org/doc/numpy/reference/generated/numpy.unravel_index.html
http://docs.scipy.org/doc/numpy/reference/generated/numpy.ravel_multi_index.html#numpy.ravel_multi_index

nathanfranklin:
> From my review It looks that the there are a number of changes related to the current CUDS description. It is not clear why they were necessary, maybe @nathanfranklin can understand it better. I am also a little confused about the use of IndexDataContainerTable, I made some suggestions but I might be missing something.

I am not aware of anything like the following changes to ABCLattice:
```
def add_lattice(self, lattice, record=None):
```

simphony/cuds/lattice.py
tuopuu: BaseLattice is fine. We can use it for now, and if we can come up with something else in future then it's easy to change.

simphony/cuds/lattice.py
tuopuu: I think it's necessary to implement these so that BaseLattice class can be instantiated. Instantiation is needed in using H5CUDS.add_lattice with a BaseLattice parameter that has all the "metadata" included, but no DataContainer data. Also, we think that BaseLattice will be used as a common ancestor for all future XLattice classes. 

In the case of FileLattice, typical lattice data (used in actual simulation) does not fit into memory of a single pc. However, it is useful to allow adding of a FileLattice to a H5CUDS-file by using just a single computer instead of a computer cluster where the full lattice data could fit into. Using a common ancestor here keeps the required parameters add_lattice nice and clean, and it still works as before.

simphony/io/file_lattice.py
tuopuu: You are right. self._table[n] does the same job in much cleaner way.

simphony/cuds/lattice.py
kemattil: First of all, I think that the ABCLattice interface should include

@abstractproperty
    def name(self):

or even

@property
    def name(self):

and the same for all attributes listed in the docstings of ABCLattice (see https://docs.python.org/2/library/abc.html for abstract properties and related get/set-methods). Also, ABCLattice should probably include

 def __init__(self, name, type, base_vect, size, origin):

whether it is implemented or not (depending on @abstractproperty vs @property).

@tuopuu: do we really need the MetaLattice (or BaseLattice) class if we define a "full" interface for ABCLattice?
That is, are you really instantiating MetaLattice somewhere (in order to have an empty lattice)?

simphony/io/file_lattice.py
tuopuu: You are right. Removed keywords completely. It was originally there to allow selecting of CUBA-keys that are updated. I think that we can assume that when a node needs to be updated, all the columns are updated and old data is overwritten.

simphony/io/file_lattice.py
tuopuu: node.data might have additional CUBA-keys defined with respect to FileLattice._table, so those have to filtered away before copying.

simphony/io/file_lattice.py
tuopuu: Yes, that's better. It could be faster as well.

simphony/io/file_lattice.py
itziakos: The IndexedDataContainer does the filtering for you. There should be no need to do anything else.

simphony/io/h5_cuds.py
tuopuu: We need a way of adding a lattice to a file with only a selected set of keywords, instead of all of them. This is currently done by using a custom record class that describes the Pytables columns. 

Only two solutions came to my mind how to solve this problem: 
1) Having an optional record parameter here, that's a reference to a custom IsDescription class that describes the columns and datatypes. [chosen]
2) Copying all the needed data from the original lattice to a new one, and adding that lattice to the H5CUDS-file. This gets rid of the record parameter in add_lattice, but it can multiply the memory requirements for the lattice (at the moment of creating the new one). Lattice-Boltzmann code is already run by filling the computer cluster node memory so every additional memory hungry operations should be avoided.

If you can come up with some other method of selecting the CUBA-keys, I'm interested in hearing that. :)

simphony/io/file_lattice.py
tuopuu: Ok, great! I missed that for some reason. I'll try to use that.

simphony/cuds/lattice.py
itziakos: The only logic that the BaseLattice provides is the `__init__` which I think is trivial enough to not require a new parent class. 

I am also not in favor of the BaseLattice overriding the abstract methods. I think is it not the right solution to the *Empty* Lattice problem. I better solution and one that have been discussed in the wiki refers to introducing a `create_lattice` method to the ABCLattice, so one can create lattice instances.

simphony/io/h5_cuds.py
itziakos: > We need a way of adding a lattice to a file with only a selected set of keywords, instead of all of them. This is currently done by using a custom record class that describes the Pytables columns.

What is not clear in this case is why you need to be able to define the record and what specific problem you are trying to solve. I thought that the data that are saved in a Lattice are a subset of CUBA keys that  is **fixed**. In that case there is no need to pass the `record` argument, one already knows what to use for a Lattice. However, if the list of keywords is not **fixed** then what is the advantage of having a smaller disk space footprint but every time one needs to add one more keywords, the complete Lattice instance has to be duplicated. In that case just use the full record.

To summarize, it looks that you are advocating that it is important to provide a record attribute for FileLattice in order to keep the disk space small based on the logic that this will speed up operations. The main problem is that changes like that should be first discussed with a usecase and a design proposal. This has happened in a number of small and large issues as one can find in github and in the wiki.

Note generic comment:

>2) Copying all the needed data from the original lattice to a new one, and adding that lattice to the H5CUDS-file. This gets rid of the record parameter in add_lattice, but it can multiply the memory requirements for the lattice (at the moment of creating the new one). Lattice-Boltzmann code is already run by filling the computer cluster node memory so every additional memory hungry operations should be avoided.

Independent of the fact that I like or not like this solution, it is not clear why one will need to make a full copy in memory of the Lattice when there are iterators that provide fast and memory efficient access to the containers.

tuopuu: The use-case with a problem is the following: We have a 3D Lattice, that is used to hold in-memory data of some Lattice-Boltzmann simulation and it containes for each LatticeNode CUBA-data for density, material_id, velocity, and a 19-dimensional 64bit float vector, that is used internally in the simulation engine. Now, only density, material_id and velocity are requested to be saved into a H5CUDS-file. 

Of course, this could be done using the FileLattice constructor directly with a specific custom record class, and then iterate over the Lattice and then update the FileLattice nodes. However, this would separate full and partial saving of lattices under different methods (add_lattice and FileLattice.__init__ + etc.) I hope that we can somehow allow partial adding of the CUBA-keys with add_lattice so that it would still work similarly to the other CUDS-types and their related H5CUDS-methods. I also think that partial adding, in the case of Lattice-Boltzmann, will be the more often used scenario.

itziakos: > and a 19-dimensional 64bit float vector, that is used internally in the simulation engine.

Is this related to a CUBA keywords or is part of the internal representation of a specific engine? Do all engines will need to save these values? If the float vector was part of the CUBA keywords that are saved wouldn't that remove the need for extended the table columns?

>  I hope that we can somehow allow partial adding of the CUBA-keys with add_lattice so that it would still work similarly to the other CUDS-types and their related H5CUDS-methods. 

If such a requirement is necessary, then an `extend_lattice` method would be more appropriate.

> I also think that partial adding, in the case of Lattice-Boltzmann, will be the more often used scenario.

If Table extension is taking place all the time then why do we use a restricted record?



I would suggest that the `records` attribute is not part of this PR and a SSB discussion is initiated (similar to what is taking place now with sharing) regarding the usecase and possible solutions to the issues that are raised. @nathanfranklin, @kemattil, @ahashibon what do you think.

nathanfranklin: Yes, i think that this isn't part of this PR and that the discussion should take place somewhere else.

tuopuu: "Is this related to a CUBA keywords or is part of the internal representation of a specific engine? Do all engines will need to save these values? If the float vector was part of the CUBA keywords that are saved wouldn't that remove the need for extended the table columns?"

I think we are discussing of slightly different topics here. We do not (often) have to extend a FileLattice table after its creation. Typically before the simulation, we already know what quantities are relevant, and a FileLattice can be created with empty columns for those quantities. The problem is that a Lattice can have something else stored in it (like this "nearly incompressible" 19-dim vector array used in a certain Lattice-Boltzmann model).

We can move this discussion elsewhere and I can modify the PR so that partial saving is only supported with FileLattice constructor for now.

kemattil: LB modeling engines internally store (at least) DENSITY, VELOCITY, and DISTRIBUTION (the above mentioned 19-component vector). On the other hand, at the SimPhoNy side, lattices typically need to store/hold only, for example, DENSITY and VELOCITY (usual variables of interest). So, if H5Cuds currently allows storing a subset of CUBA keywords, this case is already ok from my point of view.

However, when we proceed to more advanced couplings between methods and scales, it might become necessary/fruitful to bring the DISTRIBUTION data (the true/intrinsic variable of LBM) over to the SimPhoNy side. But even in this case, it is probably not necessary to store the distributions to a file (at least when saving intermediate time snapshots of the lattice state). Hence, a demand for storing a subset of a subset of CUBA keywords can arise.

At the moment I don't know how such cases would be best handled. But at the current development stage, and in order to proceed, we can probably leave this feature to future discussion and design.

nathanfranklin: @kemattil , @tuopuu , can you open an issue.

tuopuu: Opened the issue #70

simphony/io/h5_cuds.py
tuopuu: Changed to self._file. FileLattice constructor needs a reference to the file it uses.

simphony/io/h5_cuds.py
tuopuu: 'eval' is not needed anymore here, since add_lattice only creates full tables, so I removed it. I'll try to find another way of selecting the CUBA-keys in the case that only some of them is saved into a file.

tuopuu:
@itziakos, could you please have a look.

itziakos:
@tuopuu, the PR contains a good number of files that are already in master as new. This is a little strange and defently makes reviewing difficult. Did you use rebase (we realy should not use rebase with branches that are tracking remote repos)? Can you merge master back to this PR?

tuopuu:
Yes, I rebased from master. I can avoid that in future, but generally speaking, what is then the best way of updating out-of-date files in a branch from the current master? And could you please show how to "merge master back" in this case?

itziakos:
> Yes, I rebased from master. I can avoid that in future, but generally speaking, what is then the best way of updating out-of-date files in a branch from the current master? And could you please show how to "merge master back" in this case?

A common workflow merge new updates/changes from master and preserve the commit history when using remote repositories (i.e. github) is as follows:

```
git fetch origin  # fetch changes from the remote
git checkout master # checkout local master
git merge origin/master or git pull # update local master with remote changes
git checkout <feature branch> # checkout feature branch
git merge master # merge current master changes into the feature branch
```
The last operation (is what I call "merge master back to the feature branch"), might complain with conflicts and halt the merge operation. If conflicts appear please resolve them before committing the merge operation. 

Based on how this PR looks, It think (guess) that you re-based master on-top of your branch which resulted in re-playing (creates new hashes) the master commits on-top of your branch. As a result github will show in the PR the new re-created master commits as part of the PR. And I guess that you would have to use `--force` in order to push changes into the feature branch or maybe that was the reason why you did not continue work in the `file_lattice` branch.

Now returning back to this PR, it is not possible to review it the way it stands because it is and I would not suggest to merge it because it really messes up the commit history. I can try to fix the problem but this will not happen today.

tuopuu:
Thank you, @itziakos. While learning to use git, I could just create a new branch from master and copy+add the files related to FileLattice manually, so that we can get rid of the mess.

itziakos:
I would suggest to start with the `file_lattice` branch (create a file_lattice_clean one for that) merge the current master into the branch (or maybe you can do this last) and cherry pick ***your*** the commits one after the other from the `file_lattice2` into the file_lattice_new (This is what I would do). As long as you cherry pick your commits in order you will have the list amount of problems. 

I usually use a graphical interface when cherry-picking and I find gitk (it should be available by default) to be very powerfull tool for such operations.

tuopuu:
Created a new pull request #74 which replaces this. Please, have a look when you have time.

PR #74, opened 2015-03-03
PR: file_lattice_clean

tuopuu:
Picked the commits from file_lattice2 into a new clean branch. This pull request adds basic functionality for H5CUDS-based lattices.
- H5CUDS now has add/get/delete/iter_lattices methods
- FileLattice implements ABCLattice in a H5CUDS-file
- unittests for new methods are included

### DIFF
--- a/simphony/cuds/lattice.py
+++ b/simphony/cuds/lattice.py
@@ -11,7 +11,7 @@ class LatticeNode:
     Attributes
     ----------
     index : tuple of D x int
-        Node index coordinate
+        D dimensional node index coordinate
     data : DataContainer
 
     """
@@ -31,14 +31,15 @@ class Lattice(ABCLattice):
 
     Attributes:
     -----------
-    name : string
+    name : str
         name of the lattice
-    type : string
+    type : str
         Bravais lattice type (should agree with the base_vect below).
     base_vect : D x float
-        defines a Bravais lattice (an alternative for primitive vectors).
+        defines a Bravais lattice of dimension D = 2,3
+        (an alternative for primitive vectors).
     size : tuple of D x int
-        number of lattice nodes (in the direction of each axis).
+        number of lattice nodes in the direction of each axis.
     origin : D x float
         origin of lattice
 
--- a/simphony/cuds/lattice.py
+++ b/simphony/cuds/lattice.py
@@ -14,7 +14,6 @@ class LatticeNode:
         Node index coordinate
     data : DataContainer
 
-
     """
     def __init__(self, index, data=None):
         self.index = tuple(index)
@@ -26,22 +25,25 @@ class LatticeNode:
 
 
 class Lattice(ABCLattice):
-    """
-    A Bravais lattice;
-    stores references to data containers (node related data).
+    """A Bravais lattice
 
-    Parameters
-    ----------
-    name : str
-    type : str
+    Stores references to data containers (node related data).
+
+    Attributes:
+    -----------
+    name : string
+        name of the lattice
+    type : string
         Bravais lattice type (should agree with the base_vect below).
     base_vect : D x float
         defines a Bravais lattice (an alternative for primitive vectors).
-    size : tuple of D x size
+    size : tuple of D x int
         number of lattice nodes (in the direction of each axis).
     origin : D x float
+        origin of lattice
 
     """
+
     def __init__(self, name, type, base_vect, size, origin):
         self.name = name
         self._type = type
@@ -50,28 +52,12 @@ class Lattice(ABCLattice):
         self._origin = np.array(origin, dtype=np.float)
         self._dcs = np.empty(size, dtype=object)
 
-    @property
-    def type(self):
-        return self._type
-
-    @property
-    def base_vect(self):
-        return self._base_vect
-
-    @property
-    def size(self):
-        return self._size
-
-    @property
-    def origin(self):
-        return self._origin
-
     def get_node(self, index):
         """Get a copy of the node corresponding to the given index.
 
         Parameters
         ----------
-        index: tuple of D x int (node index coordinate)
+        index : tuple of D x int (node index coordinate)
 
         Returns
         -------
@@ -86,7 +72,7 @@ class Lattice(ABCLattice):
 
         Parameters
         ----------
-        lat_node: reference to a LatticeNode object
+        lat_node : reference to a LatticeNode object
             data copied from the given node
 
         """
@@ -98,7 +84,7 @@ class Lattice(ABCLattice):
 
         Parameters
         ----------
-        indices: iterable set of D x int, optional
+        indices : iterable set of D x int, optional
             node index coordinates
 
         Returns
@@ -127,18 +113,34 @@ class Lattice(ABCLattice):
         """
         return self.origin + self.base_vect*np.array(index)
 
+    @property
+    def type(self):
+        return self._type
+
+    @property
+    def base_vect(self):
+        return self._base_vect
+
+    @property
+    def size(self):
+        return self._size
+
+    @property
+    def origin(self):
+        return self._origin
+
 
 def make_hexagonal_lattice(name, h, size, origin=(0, 0)):
     """Create and return a 2D hexagonal lattice.
 
     Parameters
     ----------
-    name: str
-    h: float
+    name : str
+    h : float
         lattice spacing.
-    size: 2 x int
+    size : 2 x int
         number of lattice nodes (in each axis direction).
-    origin: 2 x float (default value = (0,0))
+    origin : 2 x float (default value = (0,0))
         lattice origin.
 
     Returns
@@ -155,12 +157,12 @@ def make_square_lattice(name, h, size, origin=(0, 0)):
 
     Parameters
     ----------
-    name: str
-    h: float
+    name : str
+    h : float
         lattice spacing.
-    size: 2 x int
+    size : 2 x int
         number of lattice nodes (in each axis direction).
-    origin: 2 x float (default value = (0,0))
+    origin : 2 x float (default value = (0,0))
         lattice origin.
 
     Returns
--- a/simphony/io/file_lattice.py
+++ b/simphony/io/file_lattice.py
@@ -43,7 +43,7 @@ class FileLattice(ABCLattice):
 
         # If Lattice not in file, create a lattice
         if self._name not in self._file.root.lattice:
-            if None in [type, base_vect, size, origin]:
+            if type is None:
                 error_str = ("Lattice '{}' does not exist in file. "
                              "Type, base_vect, size and origin must "
                              "be given proper values.")
--- a/simphony/io/file_lattice.py
+++ b/simphony/io/file_lattice.py
@@ -36,7 +36,6 @@ class FileLattice(ABCLattice):
     """
     def __init__(self, file, name, type, base_vect, size, origin,
                  record=None):
-        self.name = name
         self._type = type
         self._base_vect = np.array(base_vect, dtype=np.float)
         self._size = tuple(size)
@@ -46,7 +45,7 @@ class FileLattice(ABCLattice):
         self._record = record
 
         # If Lattice not in file, create a lattice
-        if self.name not in self._file.root.lattice:
+        if name not in self._file.root.lattice:
             if self._record is None:
                 # If record not specified use NoUIDRecord
                 self._record = NoUIDRecord
@@ -64,7 +63,7 @@ class FileLattice(ABCLattice):
 
         # If Lattice already in file, return a reference
         else:
-            self._table = IndexedDataContainerTable(self._group, self.name)
+            self._table = IndexedDataContainerTable(self._group, name)
             lattice = self._group._f_getChild(self.name)
             self._type = lattice.attrs.type
             self._base_vect = lattice.attrs.base_vect
@@ -148,3 +147,11 @@ class FileLattice(ABCLattice):
     @property
     def origin(self):
         return self._origin
+
+    @property
+    def name(self):
+        return self._table._table._v_name
+
+    @name.setter
+    def name(self, value):
+        self._table._table._f_rename(value)
--- a/simphony/io/file_lattice.py
+++ b/simphony/io/file_lattice.py
@@ -1,0 +1,150 @@
+from simphony.cuds.lattice import ABCLattice
+from simphony.cuds.lattice import LatticeNode
+from simphony.io.indexed_data_container_table import IndexedDataContainerTable
+from simphony.io.data_container_description import NoUIDRecord
+from simphony.core.data_container import DataContainer
+
+import numpy as np
+
+
+class FileLattice(ABCLattice):
+    """
+    FileLattice object to use in file lattices.
+
+    Attributes
+    ----------
+    file : PyTables file
+        reference to PyTables file where the lattice is or will be located
+    name : str
+        name of the lattice
+    type : string
+        Bravais lattice type (should agree with the _base_vect below).
+    base_vect : D x float
+        defines a Bravais lattice (an alternative for primitive vectors).
+    size : D x size
+        number of lattice nodes (in the direction of each axis).
+    origin : D x float
+        origin of lattice
+    record : tables.IsDescription
+        A class that describes column types for PyTables table.
+
+    Returns:
+    ----------
+    FileLattice
+        The lattice newly added to the file or existing in the file.
+
+    """
+    def __init__(self, file, name, type, base_vect, size, origin,
+                 record=None):
+        self.name = name
+        self._type = type
+        self._base_vect = np.array(base_vect, dtype=np.float)
+        self._size = tuple(size)
+        self._origin = np.array(origin, dtype=np.float)
+        self._file = file
+        self._group = file.root.lattice
+        self._record = record
+
+        # If Lattice not in file, create a lattice
+        if self.name not in self._file.root.lattice:
+            if self._record is None:
+                # If record not specified use NoUIDRecord
+                self._record = NoUIDRecord
+            self._table = IndexedDataContainerTable(
+                self._group, name, self._record, np.prod(self._size))
+            for i in xrange(np.prod(self._size)):
+                self._table.append(DataContainer())
+
+            # Set table attributes
+            lattice = self._group._f_getChild(self.name)
+            lattice.attrs.type = self._type
+            lattice.attrs.base_vect = self._base_vect
+            lattice.attrs.size = self._size
+            lattice.attrs.origin = self._origin
+
+        # If Lattice already in file, return a reference
+        else:
+            self._table = IndexedDataContainerTable(self._group, self.name)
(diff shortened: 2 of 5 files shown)


simphony/io/tests/test_h5_cuds.py
```diff
@@ -346,6 +347,40 @@ def test_mesh_rename(self):
...
+        lat2 = Lattice('test_lattice2', 'cubic', (1.0, 1.0, 1.0),
+                       (3, 3, 3), (0.0, 0.0, 0.0))
+
+        filename = os.path.join(self.temp_dir, 'test.cuds')
+        with closing(H5CUDS.open(filename, 'w')) as handle:
+            handle.add_lattice(lat)
+            handle.add_lattice(lat2)
+            for lat in handle.iter_lattices():
+                self.assertTrue(lat.name in ['test_lattice', 'test_lattice2'])
+
```
nathanfranklin: Could a test of when the lattice is renamed also be added. 

See test_mesh_rename and test_particle_container_rename

simphony/cuds/lattice.py
```diff
@@ -26,22 +25,25 @@ def __init__(self, index, data=None):
...
 
-    Parameters
-    ----------
-    name : str
-    type : str
+    Stores references to data containers (node related data).
+
+    Attributes:
+    -----------
+    name : string
```
itziakos: `str` is better than `string` since `str` is the name of the string type in python

simphony/cuds/lattice.py
```diff
@@ -26,22 +25,25 @@ def __init__(self, index, data=None):
...
+        name of the lattice
+    type : string
         Bravais lattice type (should agree with the base_vect below).
     base_vect : D x float
         defines a Bravais lattice (an alternative for primitive vectors).
-    size : tuple of D x size
+    size : tuple of D x int
```
itziakos: In the docstrings it is not explained what is `D`

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,150 @@
...
+        Parameters:
+        -----------
+        index : tuple of D x int (node index coordinate)
+
+        Returns:
+        -----------
+        A reference to a LatticeNode object
+
+        """
+        n = np.ravel_multi_index(index, self._size)
```
itziakos: I was not sure if these methods where doing the right thing. Thanks for looking up

simphony/io/h5_cuds.py
```diff
@@ -208,6 +262,21 @@ def delete_mesh(self, name):
...
+        """Delete lattice from file.
+
+        Parameters
+        ----------
+        name : str
+            name of lattice to delete
+        """
+        try:
+            filelat = self._handle.root.lattice._f_get_child(name)
+            filelat._f_remove(recursive=True)
```
itziakos: I would move this line out of the try... except block, it is prefered to have only the code that would raise an exception in the try block. Consider the alternative
```python
try:
    filelat = self._handle.root.lattice._f_get_child(name)
except tables.NoSuchNodeError:
    raise ValueError('Lattice \'{n}\` does not exist'.format(n=name))
else:
    filelat._f_remove(recursive=True)
```

simphony/io/tests/test_h5_cuds.py
tuopuu: Added a new commit that implements/fixes this functionality.

simphony/cuds/lattice.py
tuopuu: OK, I'll use that in future.

simphony/io/tests/test_file_lattice.py
```diff
@@ -0,0 +1,180 @@
...
+        shutil.rmtree(self.temp_dir)
+
+    def test_file_lattice_constructed_from_Lattice(self):
+        """ Checks that FileLattice constructed from Lattice with a
+        CustomRecord column description has correct attributes
+
+        """
+        self.assertEqual(self.filelattice.name, 'test_lattice')
+        self.assertEqual(self.filelattice.type, 'cubic')
+        self.assertItemsEqual(self.filelattice.base_vect,
```
itziakos: These should probably be using assertSequenceEqual. But I think that using the `assert_array_equal` from numpy tests is probably the most appropriate method here.
you can import the method as
```python
from numpy.testing import assert_array_equal
```

simphony/cuds/lattice.py
tuopuu: I'll add a note on the accepted values of D.

simphony/io/tests/test_file_lattice.py
```diff
@@ -0,0 +1,180 @@
...
+        the nodes is returned when the function iter_nodes is called
+        without arguments
+
+        """
+        fl_nodes = self.filelattice.iter_nodes()
+
+        for M in fl_nodes:
+            N = self.lattice.get_node(M.index)
+            self.assertEqual(N.index, M.index)
+            self.assertItemsEqual(N.data, M.data)
```
itziakos: This will compare only the keys in the `DataContainers`. I would suggest to use the utility methods in https://github.com/simphony/simphony-common/blob/master/simphony/testing/utils.py#L26

simphony/io/tests/test_file_lattice.py
```diff
@@ -0,0 +1,180 @@
...
+        """ Checks the node iterator on a subset of nodes
+
+        """
+
+        fl_nodes = self.filelattice.iter_nodes([(0, 0, 0), (0, 1, 2)])
+
+        for M in fl_nodes:
+            N = self.lattice.get_node(M.index)
+            self.assertEqual(N.index, M.index)
+            self.assertItemsEqual(N.data, M.data)
```
itziakos: See comment in https://github.com/simphony/simphony-common/pull/74/files#r25767404

simphony/io/tests/test_file_lattice.py
```diff
@@ -0,0 +1,180 @@
...
+                        {CUBA.MATERIAL_ID: 2, CUBA.DENSITY: 10.0,
+                         CUBA.VELOCITY: (0.0, 0.0, 10.0)})
+
+        self.filelattice.update_node(N)
+
+        M = self.filelattice.get_node((3, 2, 3))
+
+        self.assertItemsEqual(N.index, M.index)
+
+        for key, value in M.data.iteritems():
```
itziakos: see comment See comment in https://github.com/simphony/simphony-common/pull/74/files#r25767404

simphony/io/tests/test_h5_cuds.py
```diff
@@ -346,6 +347,67 @@ def test_mesh_rename(self):
...
+        lat2 = Lattice('test_lattice2', 'cubic', (1.0, 1.0, 1.0),
+                       (3, 3, 3), (0.0, 0.0, 0.0))
+
+        filename = os.path.join(self.temp_dir, 'test.cuds')
+        with closing(H5CUDS.open(filename, 'w')) as handle:
+            handle.add_lattice(lat)
+            handle.add_lattice(lat2)
+            for lat in handle.iter_lattices():
+                self.assertTrue(lat.name in ['test_lattice', 'test_lattice2'])
```
itziakos: This logic will not fail iif for some reason there are two lattices with the same name or only one lattice.
I would suggest the following
```python
stored_names = [lattice.name for lattice in handle.iter_lattices()]
self.assertItemsEqual(stored_names, ['test_lattice', 'test_lattice2'])
```

simphony/io/tests/test_h5_cuds.py
```diff
@@ -346,6 +347,67 @@ def test_mesh_rename(self):
...
+                      (2, 3, 4), (0.0, 0.0, 0.0))
+
+        filename = os.path.join(self.temp_dir, 'test.cuds')
+        with closing(H5CUDS.open(filename, 'w')) as handle:
+            handle.add_lattice(lat)
+            lat2 = handle.get_lattice('test_lattice')
+
+            self.assertEqual(lat2.name, 'test_lattice')
+
+            for N in lat2.iter_nodes():
```
itziakos: See comment in https://github.com/simphony/simphony-common/pull/74/files#r25767404

simphony/cuds/lattice.py
itziakos: For consistancy please revert the change back to  `str`.

simphony/io/tests/test_file_lattice.py
tuopuu: Thanks for the tip. I'm now using assert_array_equal (without the s) for all numpy array checks.

simphony/io/tests/test_file_lattice.py
tuopuu: I was not aware that only keys are compared in this case. However, I couldn't make compare_data_containers work: it gives out an error "NoneType has no attribute assertEqual" on the line https://github.com/simphony/simphony-common/blob/master/simphony/testing/utils.py#L28

I don't know where does this error come from. But to move on, I'll just add manual comparison to the data with numpy.testing.assert_equal, which seems to work fine.

simphony/io/tests/test_h5_cuds.py
tuopuu: You are right. I updated the code with your lines.

simphony/io/tests/test_file_lattice.py
itziakos: The compare function needs the test-case to be passed as a keyword argument. 
The simple way is to pass `self` as the testcase argument or register the method with the unittest machinery as in the example https://github.com/simphony/simphony-common/blob/4ad04227e7bbbeaf61f9b458eced5701b3584072/simphony/testing/abc_container_check.py#L15

simphony/io/h5_cuds.py
```diff
@@ -177,6 +209,28 @@ def get_mesh(self, name):
...
+        file has been closed then the lattice should
+        no longer be used.
+
+        Parameters
+        ----------
+        name : str
+            name of lattice to return
+        """
+        if name in self._handle.root.lattice:
+            F = self._handle.root.lattice._v_leaves[name].attrs
```
itziakos: I think this is not necessary since the ``__init__`` method of FileLattice will recover the right values.

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,157 @@
...
+    record : tables.IsDescription
+        A class that describes column types for PyTables table.
+
+    Returns:
+    ----------
+    FileLattice
+        The lattice newly added to the file or existing in the file.
+
+    """
+    def __init__(self, file, name, type, base_vect, size, origin,
```
itziakos: type, base_vect, size and origin should be optional since they are recovered from the hdf5 File if the lattice already exists.

simphony/io/h5_cuds.py
itziakos: This ofcourse assumes that https://github.com/simphony/simphony-common/pull/74/files#r26033012 has the arguments optional.

simphony/io/file_lattice.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        if indices is None:
+            for row_number, data in enumerate(self._table):
+                index = np.unravel_index(row_number, self._size)
+                yield LatticeNode(index, data)
+        else:
+            for index in indices:
+                yield self.get_node(index)
+
+    def get_coordinate(self, index):
```
itziakos: It does not look that there are tests for this method

itziakos:
@tuopuu, I had another look with only minor comment. Thanks for cleaning up the branch.

simphony/io/file_lattice.py
tuopuu: Thank you, that's a good point. I'll change this and the related function calls in tests and h5_cuds.get_lattice.

simphony/io/file_lattice.py
tuopuu: Thank you, I'll add it.

itziakos: Please make sure you add some tests for the `OrthotombicP` and `Hexagonal` lattice

tuopuu:
@itziakos, could please take another look. I modified FileLattice constructor slightly adding None defaults to the attributes, but this creates a FutureWarning. My intention is to print out an error, if FileLattice is called with a name not existing in file, and without the optional arguments. Could you think of any other way of defining the optional arguments, or making the validity check? In my opinion, using some real values there (like,'Cubic',(1.0,1.0,1.0), etc.) would not look elegant at all. What do you think?

nathanfranklin:
@roigcarlo , @SGGgarcia  can you all take a quick review too (time permitted).

itziakos:
>  I modified FileLattice constructor slightly adding None defaults to the attributes, but this creates a FutureWarning, 

What is the FutureWarning, can you provide a trace back?

SGGgarcia:
I took a quick view and looks ok :+1:

nathanfranklin:
> I do not think that you need to check that these arguments are None. If the name is not there one should create a Lattice using default values.

It seems FileLattice is to be used when there is a lattice in a file (`get_lattice` in `H5CUDS`) or a existing lattice is to be added to a file (`add_lattice` in `H5CUDS`).

Maybe to simplify your ctor of FileLattice you could just have:
```    def __init__(self, file, name=None, lattice=None):``` 
instead of
```    def __init__(self, file, name, type=None, base_vect=None, size=None, origin=None, record=None):```

Then like @itziakos  said, just check if `name` is not there and then you can query the `lattice` parameter to setup your lattice stored in the file.  If you do make such changes, I would also move the original 'updating' of nodes that you perform in `add_lattice` there.

tuopuu:
@itziakos The FutureWarning is this: "FutureWarning: comparison to 'None' will result in an elementwise object comparison in the future." 
It refers to line 46 in file_lattice.py: "if None in [type, base_vect, size, origin]:"

tuopuu:
@nathanfranklin, you're right, FileLattice constructor functionality overlaps with H5CUDS get_lattice/add_lattice in some ways. However, I think it's inevitable because we want to be able to create an in-file lattice without creating it fully in memory.

> Maybe to simplify your ctor of FileLattice you could just have:
> def __init__(self, file, name=None, lattice=None):

For lattice systems we currently always have a full numpy.ndarray (of at least empty elements, possibly with real CUBA-data) generated which allocates memory even if it is empty. Therefore first generating an empty Lattice that is used in construction of FileLattice is not very efficient, and in the case of big lattices, would possibly use more memory that is currently available. In the earlier stages of this PR I introduced a new class (BaseLattice) between ABCLattice and Lattice/FileLattice, that would have just the lattice attributes defined. With a BaseLattice using the constructor you suggested would work fine! However, it was suggested by @itziakos that using such in-between class should not be part of this PR, so we had to go back to passing the lattice attributes to FileLattice separately in order to have the functionality we need.

Just to summarize, we need from the constructor to
- give a reference to an existing lattice in H5CUDS-file if the name matches. In this case optional arguments can be undefined or given value None, because they are not used.
- create an empty lattice (with full pytables table) in H5CUDS-file if the name doesn't match, use optional parameters to describe it's shape, Record, etc. In this case, all optional arguments except the Record must be given, or it will result to an error message.
- not copy contents of an existing Lattice to H5CUDS-file: this is done by H5CUDS.add_lattice and FileLattice.update_node

I think a correct way to proceed is just to get rid of the FutureWarning message for now. I'll try to fix it somehow.

itziakos:
> @itziakos The FutureWarning is this: "FutureWarning: comparison to 'None' will result in an elementwise object comparison in the future." 
> It refers to line 46 in file_lattice.py: "if None in [type, base_vect, size, origin]:"

Can you try using a tuple instead of a list?

itziakos:
> I think a correct way to proceed is just to get rid of the FutureWarning message for now. I'll try to fix it somehow.

I agree with this statement.

tuopuu:
> Can you try using a tuple instead of a list?

Unfortunately, using a tuple does not seem to help. I think comparing only the value of type is enough, because it's the first optional argument. I'll make a new commit shortly.

itziakos:
> With a BaseLattice using the constructor you suggested would work fine! However, it was suggested by @itziakos that using such in-between class should not be part of this PR, so we had to go back to passing the lattice attributes to FileLattice separately in order to have the functionality we need.

Just to clarify, there are two problems that interact here:

- The current CUDS api does not allow creating new CUDS containers from the ABCEngine class. Touching this issue there are a number of discussion in the simphoni wiki and we should push for a more formal discussion/proposal in the wiki for the SSB to evalute.
-  The Lattice implementation preallocates space which, in combination with the current api limitation can lead to memory problems. There are a number of solutions for Lattice initialization issue, but I will suggest to wait since there are a number of changes in discussion for the api like `shared_data`, that will dictate better which solution is more appropriate.

Using a BaseLattice was not a solution, but a workaround which at the same time introduced other problems like overriding the ABCMeta restrictions of ABCLattice and introducing a partially implemented class.

nathanfranklin:
> @nathanfranklin, you're right, FileLattice constructor functionality overlaps with H5CUDS get_lattice/add_lattice in some ways. However, I think it's inevitable because we want to be able to create an in-file lattice without creating it fully in memory.
>> Maybe to simplify your ctor of FileLattice you could just have:
>> def __init__(self, file, name=None, lattice=None):

> For lattice systems we currently always have a full numpy.ndarray (of at least empty elements, possibly with real CUBA-data) generated which allocates memory even if it is empty. **Therefore first generating an empty Lattice that is used in construction of FileLattice is not very efficient**, and in the case of big lattices, ...

*bold added by me

In my suggestion (``def __init__(self, file, name=None, lattice=None)`` in FileLattice), the only use of this __init__ of FileLattice where the `lattice` parameter would not equal `None` would be in the 'add_lattice' method in `H5CUDS`.  As this lattice already exists (as it was an argument in `add_lattice`, I don’t understand the concern about memory with my specific suggestion. 

I do see my suggestion forgot the ``record`` parameter.

Regardless, my suggestion can be ignored as its not important to closing this PR.

kemattil:
Currently the standalone Lattice implementation indeed preallocates memory (numpy array of references to DataContainers, references initially = None). However, this is totally internal to the standalone Lattice implementation. If empty lattices are needed, you can change the implementation: postpone the memory allocation until first update to node data is executed.
Would this solve the problem?

tuopuu:
> If empty lattices are needed, you can change the implementation: postpone the memory allocation until first update to node data is executed.

But this would require adding to lattice.update_node something like
if self._dc is not None:
   self._dc = np.empty(size, dtype=object)

which would be checked every time a node is updated. That could mean possibly billions of if checks for large lattices. One option would perhaps be to have two versions of update_node, one with the if check and one without, and use a reference/pointer to the correct function when calling update_node based on whether the array exist or not. Although I don't know if this is possible in Python.

itziakos:
The PR looks ok to me, but we should to revisit the hdf5 backend and how containers are initialized.

tuopuu:
So, this is OK to merge? We can create separate PR's that fix other issues, including the ones you mentioned.

itziakos:
yup

tuopuu:
Merged #74

PR #87, opened 2015-03-10
Issue rename abcparticlecontainer, data attribute for particles

SGGgarcia:
Hi,
this PR fixes:
    - Rename of particlecontainer to particles in all required files
    - The data attribute of Particles now is private and accessed as a property, making a copy of the Datacontainer.

pd.: never mind about the change of the name of particlesclasses.py -> particles.py, that was already covered.


nathanfranklin:
Looks good.  There are a few uses of ABCParticleContainer in the documentation. See the results of `grep -RnE '(ParticleContainer|particle_container)' . `  This PR should also update the particle-related methods in `ABCModelingEngine` and `H5Cuds` (i.e.  `xxx_particle_container` to `xxx_particles` )

simphony/cuds/particles.py
```diff
@@ -448,6 +449,19 @@ def _update_element(self, cur_dict, element, clone):
         else:
             raise ValueError('id: {} does not exist'.format(uid))
 
+# ================================================================
+
+    # properties
+
+# ================================================================
+    def _get_data(self):
+        return copy.deepcopy(self._data)
```
itziakos: I think using deepcopy here is probably an overkill. Using `DataContainer(self._data)` should be enough. There is only one possible failure mode when the `values` in the dictionary are mutable objects like list or numpy arrays. However, I thought that we do not use them (if we do then it should probably be an error).

itziakos: Please note that the current code is not wrong and my comment does not have to be resolved in this PR.

SGGgarcia:
@itziakos good point, I can change it now anyway
@nathanfranklin I missed that, thanks; just a minor comment, the 'iter_particle_containers' would be 'iter_particles', that has the same name as the 'iter_particles' in the ABCParticles class. Maybe someone thinks that it looks a little bit weird, any problems with this?

simphony/cuds/particles.py
```diff
@@ -448,6 +449,19 @@ def _update_element(self, cur_dict, element, clone):
...
+
+    def _set_data(self, new_data):
+        self._data = copy.deepcopy(new_data)
+
+    data = property(_get_data, _set_data)
```
itziakos: We are missing some tests for this functionality

simphony/cuds/tests/test_particles.py
```diff
@@ -61,29 +61,39 @@ class TestNativeContainerAddParticles(
...
 
+class TestParticlesDataContainer(unittest.TestCase):
+
+    def test_data(self):
+        pc = Particles(name='foo')
+        data = pc.data
+        data[CUBA.MASS] = 9
+        pc.data = data
+        ret_data = pc.data
+        self.assertItemsEqual(data, ret_data)
```
itziakos: Please see related comments in https://github.com/simphony/simphony-common/pull/97#discussion_r26293247

itziakos:
> 'iter_particles', that has the same name as the 'iter_particles' in the ABCParticles class. Maybe someone thinks that it looks a little bit weird, any problems with this?

I also think it is weird, I personally would suggest against it since it can be confusing, but I do not mind it in this PR. Naming in general needs to be revisited as soon as the API starts to settle down.

itziakos:
@SGGgarcia, on minor comment on the tests, apart from that it looks ok.

itziakos:
It looks that there is a conflict with the current master. @SGGgarcia please resolver and push the changes

Issue #89, opened 2015-03-10
FileMesh iterators do not return data attribute


roigcarlo:
The data attribute is not being returned while iterating over all points/edges/faces/cells of a FileMesh

roigcarlo:
Solved in https://github.com/simphony/simphony-common/pull/92

PR #92, opened 2015-03-10
Bug mesh io iterator data

roigcarlo:
This PR fixes should fix https://github.com/simphony/simphony-common/issues/89, a problem reported by @khiltunen which was making the data of the entities of the filemesh not being returned if the iter_xxxx() were called. 

TODO:
- [x] Travis pass

Once travis passes I will ping you for the merge.

### DIFF
--- a/simphony/io/tests/test_file_mesh.py
+++ b/simphony/io/tests/test_file_mesh.py
@@ -382,6 +382,28 @@ class TestFileMesh(unittest.TestCase):
 
         self.assertItemsEqual(icells_id, cuids)
 
+    def test_get_all_points_iterator_with_data(self):
+        """ Checks the point iterator
+
+        Checks that an interator over all
+        the points of the mesh is returned
+        when the function iter_points is called
+        without arguments and its data is correctly retrieved
+
+        """
+
+        puids = [self.mesh.add_point(point) for point in self.points[:3]]
+
+        ipoints = self.mesh.iter_points()
+
+        ipoints_id = [point.uid for point in ipoints]
+        first_point = self.mesh.iter_points().next()
+
+        self.assertItemsEqual(ipoints_id, puids)
+        self.assertIsNot(len(first_point.data), 0)
+
+        self.assertEqual(first_point.data, DataContainer(VELOCITY=(0, 0, 0)))
+
     def test_get_all_edges_iterator_with_data(self):
         """ Checks the edge iterator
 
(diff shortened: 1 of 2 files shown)


roigcarlo:
@khiltunen, please confirm that everything is ok so we can merge

simphony/io/tests/test_file_mesh.py
```diff
@@ -375,6 +375,93 @@ def test_get_all_cells_iterator(self):
...
+
+        euids = [self.mesh.add_edge(edge) for edge in edges]
+
+        iedges = self.mesh.iter_edges()
+
+        iedges_id = [edge.uid for edge in iedges]
+        first_edge = self.mesh.iter_edges().next()
+
+        self.assertItemsEqual(iedges_id, euids)
+        self.assertIsNot(len(first_edge.data), 0)
```
itziakos: The `compare_data_containers` function can be used to setup an equality assertion for the TestCase (see usage in https://github.com/simphony/simphony-common/blob/master/simphony/testing/abc_check_particle_containers.py#L66)

Then one can just do 
```python

self.assertEqual(first_edge.data, DataContainer(VELOCITY=(0, 0, 0)))

File: simphony/io/tests/test_file_mesh.py [hunk 2]
```diff
@@ -375,6 +375,93 @@ def test_get_all_cells_iterator(self):
...
+        the edges of the mesh is returned
+        when the function iter_edges is called
+        without arguments and its data is correctly retrieved
+
+        """
+
+        puids = [self.mesh.add_point(point) for point in self.points[:3]]
+
+        edges = [
+            Edge(puids[0:2], data=DataContainer({CUBA.VELOCITY: [0, 0, 0]}))
```
itziakos: I would suggest to use immutable objects like tuples for the `DataContainer` values. Especially tuples are faster to create and take much less memory resources than lists.

simphony/io/tests/test_file_mesh.py
itziakos: the correct link is https://github.com/simphony/simphony-common/blob/master/simphony/testing/abc_container_check.py#L15

khiltunen:
It is OK on my side to merge. Fix made things work.

nathanfranklin:
Looks good.  

Can a `test_get_all_points_iterator_with_data` also be written?

itziakos:
:+1:

nathanfranklin:
:+1:

Issue #83, opened 2015-03-09
Mesh implementation does not properly support the `data` attribute.


itziakos:
The Mesh implementation does not implement the `data` attribute properly.

- The `data` attribute should be set to a `DataContainer` not `0` at initialization
- The `data` attribute should be a property so setting and getting the value should return a copy of the internal representation.

Issue #84, opened 2015-03-09
The native python implementation for the Particles container does not support properly the `data` attribute


itziakos:
`data` should probably be a property so that getting and setting the information take place through a copy.

nathanfranklin:
> data should probably be a property so that getting and setting the information take place through a copy.

just to clarify this for myself (and also for anyone new to this discussion),  we want to avoid the situation where a user could do the following:

```
pc = Particles(name="foo")
data=pc.data
# user changes data and has now changed the data
# of the high-level object particles 
data[CUBA.RADIUS] = 42
print(pc.data[CUBA.RADIUS])
```

So we have `data` now as a property so one would need to do:
 ```
pc = Particles(name="foo")
data = pc.data
data[CUBA.RADIUS] = 42
pc.data = data
print(pc.data[CUBA.RADIUS])
```
 
instead of the previous way:
```
# no longer allowed
pc = Particles(name="foo")
pc.data[CUBA.RADIUS] = 42
print(pc.data[CUBA.RADIUS])
```

itziakos:
Yes, this is how the file based CUDS objects behave. I still think that we might want to use ``data()`` and ``set_data()`` to make it clear that the information is a snapshot of the internal data but this should probably go through a discussion in the wiki.

nathanfranklin:
> Yes, this is how the file based CUDS objects behave.

all the high-level objects  (i.e. ABCParticles, ABCMesh, ABCLattice) should all behave this way. right? 

> I still think that we might want to use data() and set_data() to make it clear that the information is a snapshot of the internal data but this should probably go through a discussion in the wiki.

Sounds good

I am not sure but wonder if this discussion would relate to the CM, SP, BC of ABCModelingEngine (as they are *at the moment* similar and so similar behavior might be expected).

itziakos:
> I am not sure but wonder if this discussion would relate to the CM, SP, BC of ABCModelingEngine (as they are at the moment similar and so similar behavior might be expected).

They touch, but do not have to use the same solution or be solved at the same time.

itziakos:
> all the high-level objects (i.e. ABCParticles, ABCMesh, ABCLattice) should all behave this way. right?

Based on the api description yes. I created separate issues so that it will be easier to distribute the work

PR #97, opened 2015-03-12
Feature mesh data property

roigcarlo:
This should add the data property to the mesh class.

I agree with https://github.com/simphony/simphony-common/issues/84#issuecomment-78071986 data() and set_data(), but I understand that there is not yet a decision so I used a property instead.  

@itziakos , @nathanfranklin, @mehdisadeghi  Is this correct to fix https://github.com/simphony/simphony-common/issues/83 ?

### DIFF
--- a/simphony/cuds/mesh.py
+++ b/simphony/cuds/mesh.py
@@ -237,7 +237,7 @@ class Mesh(ABCMesh):
                 + "DataContainer expected."
             raise TypeError(error_str)
         self._data = value
-    
+
     def get_point(self, uuid):
         """ Returns a point with a given uuid.
 
--- a/simphony/cuds/mesh.py
+++ b/simphony/cuds/mesh.py
@@ -228,11 +228,11 @@ class Mesh(ABCMesh):
 
     @property
     def data(self):
-        return self._data
+        return dc.DataContainer(self._data)
 
     @data.setter
     def data(self, value):
-        self._data = value
+        self._data = dc.DataContainer(value)
 
     def get_point(self, uuid):
         """ Returns a point with a given uuid.
--- a/simphony/cuds/mesh.py
+++ b/simphony/cuds/mesh.py
@@ -218,13 +218,22 @@ class Mesh(ABCMesh):
 
     def __init__(self, name):
         self.name = name
-        self.data = 0
 
         self._points = {}
         self._edges = {}
         self._faces = {}
         self._cells = {}
 
+        self._data = dc.DataContainer()
+
+    @property
+    def data(self):
+        return self._data
+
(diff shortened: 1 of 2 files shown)


itziakos:
> I agree with #84 (comment) data() and set_data(), but I understand that there is not yet a decision so I used a property instead.

yes, this is currently only a suggestion and does not reflect the current CUDS api

simphony/cuds/mesh.py
```diff
@@ -218,13 +218,26 @@ class Mesh(ABCMesh):
...
 
+        self._data = dc.DataContainer()
+
+    @property
+    def data(self):
+        return self._data
+
+    @data.setter
+    def data(self, value):
+        if not isinstance(value, dc.DataContainer):
```
itziakos: I do no think that there is a need for checking the type (for example data containers coming from adapters will not necessarily be subclasses of DataContainer). As long as the `value` is a CUBA based mapping it should be good for us. I suggest instead to do the check indirectly as:
```
self._data = DataContainer(value)
```
Which will also make sure that we make a copy.

In general in pyhon specific type checking is avoided unless is absolutely essential to the operation
which will make sure that

simphony/cuds/mesh.py
```diff
@@ -218,13 +218,26 @@ class Mesh(ABCMesh):
...
         self._points = {}
         self._edges = {}
         self._faces = {}
         self._cells = {}
 
+        self._data = dc.DataContainer()
+
+    @property
+    def data(self):
+        return self._data
```
itziakos: Please return a copy of the data
```python
return DataContainer(self._data)
```

simphony/cuds/tests/test_mesh.py
```diff
@@ -680,5 +682,48 @@ def test_update_cell(self):
...
+        """
+
+        org_data = DataContainer()
+
+        org_data[CUBA.VELOCITY] = (0, 0, 0)
+
+        self.mesh.data = org_data
+        ret_data = self.mesh.data
+
+        self.assertItemsEqual(org_data, ret_data)
```
itziakos: This checks only the ``keys`` of mappings, please use the compare_data_container function as in https://github.com/simphony/simphony-common/blob/master/simphony/testing/abc_container_check.py#L15 
This will allow one to do
```
self.assertEqual(org_data, ret_data)
```

itziakos: We also need to check that 
```python
self.assertIsNot(org_data, ret_data)
```

So that we verify that they are equal but not the same instance as described in #83

nathanfranklin:
:+1:

PR #99, opened 2015-03-13
PR: NotImplementedError for get_coordinate on Hexagonal lattices.

tuopuu:
Before proper fixing of the problem, added a NotImplementedError when querying coordinates of a Hexagonal lattice nodes.

### DIFF
--- a/simphony/cuds/lattice.py
+++ b/simphony/cuds/lattice.py
@@ -112,6 +112,10 @@ class Lattice(ABCLattice):
         D x float
 
         """
+        if self._type == 'Hexagonal':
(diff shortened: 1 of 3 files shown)


tuopuu:
@roigcarlo, @itziakos, please, take a look.

roigcarlo:
:+1, but I am keeping the issue open, since I understand that this is only temporal

itziakos:
I agree, :+1: , but please do not merge until the 0.1.0 release is done (unless we want to add this fix to the 0.1.0 release)

tuopuu:
Ok, I'll merge when 0.1.0 is done.

PR #105, opened 2015-03-16
Add a testcase to check that the bench modules can be imported

itziakos:


itziakos:
@mehdisadeghi, do you have time for a review?

mehdisadeghi:
@itziakos I think its fine.

Issue #101, opened 2015-03-13
Using the ABCCheckParticles class should not require pytables to be available


itziakos:
There are a number of places where the pytables Records are used, thus currently pytables needs to be installed in order to be able to user the testing templates. However, it should be possible to avoid this.

nathanfranklin:
:+1:

itziakos:
@roigcarlo, this issue is addressed in #106, so I added it in the 0.1.1 milestone

roigcarlo:
:+1:

itziakos:
implemented in #106

Issue #79, opened 2015-03-06
Tests: Lattice tests do not set the size attribute correctly


itziakos:
The lattice tests for the lattice factory functions set the size of the lattice defined in 3D to a 2D size (see
https://github.com/simphony/simphony-common/blob/master/simphony/cuds/tests/test_lattice.py#L58)

I also think that the factory tests need to be a little more thorough (e.g. test that number of lattice nodes is the expected one).

roigcarlo:
Dear @kemattil, @tuopuu  Do you mind to take this for the release 0.2.0?

itziakos:
@roigcarlo and @tuopuu  this issues is addressed  by #106

itziakos:
fixed in #106

PR #106, opened 2015-03-18
Test template for lattice

### LINKED ISSUE #76: Add a ABCCheckLatticeContainers testing template
This new class will be similar to the `ABCCheckParticlesContainers` template class and will allow easier integration and more reliable testing of the multiple implementations (native or adapter) of the ABCLattice api.

itziakos:
This PR fixes #76, #79 and #101

in more details:
- A keywords.py module is added that provides an easy to access KEYWORDS dictionary with all the CUDS keywords information.
- A `keywords` command is added to the `cuba_generate` script to automatically generate the keywords.py module.
- A `ABCCheckLattice` template testing class is introduced to test Lattice implementations. The new class is used to test Lattice and H5Lattice (two configurations). 
- A number of bugs and minor fixes to the Lattice and H5Lattice implementation have been made (e.g. Lattice now raises IndexError when there are negative values in the index.)


itziakos:
@tuopuu, @kemattil, @nathanfranklin and @roigcarlo. This PR is ready for review

simphony/testing/abc_check_lattice.py
```diff
@@ -0,0 +1,229 @@
...
+        container = self.container
+
+        index = 2, 3, 4
+        node = container.get_node(index)
+        # Update with full DataContainer.
+        node.data = create_data_container()
+        container.update_node(node)
+
+        new_node = container.get_node(index)
+        # We expect only the supported CUBA to be stored.
```
tuopuu: What if we have a CustomRecord table with not all LBM supported CUBA keys defined? Does this test fail then?

tuopuu: Had a comment here, but removed it. It looks fine as is. Obviously supported_cuba can be defined to be what ever is needed in different cases: such as keys matching CustomRecord for the case of H5Lattice.

itziakos:
> Should we also check that the given index does not exceed the maximum allowed index (size-1)?

I think that the current tests check this (https://github.com/simphony/simphony-common/pull/106/files#diff-7b4bef42bc959e80c896129ee6647ec8R91), correct?

itziakos:
> And just to verify: in update_node the indexes are not checked because we assume that the indexes are not modified between calls to get_node and update_node?

@kemattil, Added test method for update_node too

itziakos:
@tuopuu and @kemattil, PR has been updated, please have another look.

kemattil:
Ok (took a while to realize that numpy checks the upper limit, but allows negative indices)!

itziakos:
> Ok (took a while to realize that numpy checks the upper limit, but allows negative indices)!

Yes negative index means that we start from the end.

Issue #88, opened 2015-03-10
consistent naming of classes in H5IO 


nathanfranklin:
we currently have:
H5CUDS
H5Particles
FileMesh
FileLattice (in PR)

We should rename things so they are consistent.

itziakos:
I think that the H5... naming should is preferable

nathanfranklin:
Agree.

itziakos:
the Lattice io wrappers have been updated according to this issue in #109

roigcarlo:
Mesh wrappers updated in PR https://github.com/simphony/simphony-common/pull/110

PR #110, opened 2015-03-19
Rename file mesh

roigcarlo:
This should implement the mesh part for https://github.com/simphony/simphony-common/issues/88


itziakos:
:+1:  if travis-ci likes it

Issue #15, opened 2014-10-01
HDF5 Storing the `data` of higher level objects like PartcileContainer


itziakos:
from @nathanfranklin:

How will we store the  ParticleContainer's data (or the `data` member of any high-level CUDS instances)?  Will we place those in a separate single-row table (e.g. at `/particle_container/foo/data`):

MaterialType | Label  
------------ | ------------- 
"MyType" | "my label"

Or can we just use the HDF5 attributes of the `/particle_container/foo/` Node to store the contents of the `data` DataContainer.  Here, I am unclear about the limitations of using HDF5-attributes.

Any thoughts on this part of the issue (storing ParticleContainer's `data`)?

ahashibon:
The suggestion from @nathanfranklin , "to place those in a separate single-row table (e.g. at  /particle_container/foo/data )" sounds very reasonable, it does not require converting the data to a string and then parsing it (which may also make using SimPhoNy CUDS a bit awkward for third parties). The node attributes may still be used for example to store the name, or some short textual user description, or some metadata on the SimPhoNy version, etc... but not actual data related information. Does this make sense to you too?

itziakos:
@ahashibon, I have provided an implementation proposal on this feature in https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/Saving_the_%60data%60_information_of_a_CUDS_container

mehdisadeghi:
Issues #83 and #84 have to be done first to close this issue.

itziakos:
Issues #83 and #84 refer to the native python CUDS containers implementations not the HDF5 backed containers. So there is no dependency between them. Please also note that #83 and #84 are not part of this milestone.

Sorry for the confusion.

mehdisadeghi:
The problem is that in current implementation, `H5CUDS` class has to copy values from native python particle containers to the h5 containers. Therefore problematic implementation of `data` property on the other classes will break the code. At least when I used `IndexedDataContainerTable` in my implementation it will not work unless python cuds objects provide proper `data` property of type `DataContainer`. That's why I did not implement this for Mesh and Lattice classes (in contradiction to the title of this issue, ”higher level objects“).

itziakos:
Yup, you are right, #83 and #84 are blocking you from testing the fix for the hdf5 based containers. @roigcarlo, @SGGgarcia and @nathanfranklin, I think that we need to include #83 and #84 into the milestone, what do you think?

nathanfranklin:
[...]

PR #120, opened 2015-03-26
PR: High level data for H5Lattice

tuopuu:
This commit fixes issue #15 for H5Lattice:
* High level CUDS data is now saved in a separate table named 'data'
* All Lattice related data (i.e. two separate tables) is now located in subgroup under /root/lattice/
* This implementation still uses PyTables attributes to store actual lattice metadata (size, etc.) instead of the new data-table

### DIFF
--- a/simphony/io/h5_cuds.py
+++ b/simphony/io/h5_cuds.py
@@ -169,7 +169,7 @@ class H5CUDS(object):
             h5lat.update_node(node)
 
         # Copy lattice high level data to the file
-        h5lat.data = {k: lattice.data[k] for k in lattice.data}
+        h5lat.data = lattice.data
         self._handle.flush()
 
         return h5lat
--- a/simphony/io/h5_cuds.py
+++ b/simphony/io/h5_cuds.py
@@ -168,6 +168,8 @@ class H5CUDS(object):
         for node in lattice.iter_nodes():
             h5lat.update_node(node)
 
+        # Copy lattice high level data to the file
+        h5lat.data = {k: lattice.data[k] for k in lattice.data}
         self._handle.flush()
 
         return h5lat
--- a/simphony/io/h5_cuds.py
+++ b/simphony/io/h5_cuds.py
@@ -158,9 +158,10 @@ class H5CUDS(object):
             raise ValueError(
                 'Lattice \'{n}\` already exists'.format(n=lattice.name))
 
-        # Create a h5lattice with all CUBA-keys defined
-        h5lat = H5Lattice.create_new(self._root.lattice, lattice.name,
-                                     lattice.type, lattice.base_vect,
+        group = self._handle.create_group('/lattice/', lattice.name)
+
+        # Create a h5_lattice with all CUBA-keys defined
+        h5lat = H5Lattice.create_new(group, lattice.type, lattice.base_vect,
                                      lattice.size, lattice.origin)
 
         # Copy the contents of the lattice to the file
@@ -224,10 +225,11 @@ class H5CUDS(object):
         name : str
             name of lattice to return
         """
-        if name in self._root.lattice:
-            lat = H5Lattice(self._root.lattice, name)
-            return lat
-        else:
+        try:
+            group = self._root.lattice._f_get_child(name)
+            h5lat = H5Lattice(group)
+            return h5lat
+        except tables.NoSuchNodeError:
             raise ValueError(
                 'Lattice \'{n}\` does not exist'.format(n=name))
 
@@ -328,7 +330,7 @@ class H5CUDS(object):
         """
         if names is None:
             for lattice in self._root.lattice._f_iter_nodes():
-                yield self.get_lattice(lattice.name)
+                yield self.get_lattice(lattice._v_name)
         else:
             for name in names:
                 yield self.get_lattice(name)
--- a/simphony/io/tests/test_h5_lattice.py
+++ b/simphony/io/tests/test_h5_lattice.py
@@ -8,6 +8,7 @@ from simphony.io.h5_lattice import H5Lattice
 from numpy.testing import assert_array_equal
 from simphony.core.cuba import CUBA
 from simphony.testing.abc_check_lattice import ABCCheckLattice
+from simphony.core.data_container import DataContainer
 
 
 class CustomRecord(tables.IsDescription):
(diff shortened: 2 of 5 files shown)


tuopuu:
@itziakos, could you please take a look on this one as well?

kemattil:
None of the lattice properties (size, origin, data, etc.) are properties of the ABCLattice.
That is, should they be defined as abstract properties in ABCLattice (@nathanfranklin , @itziakos, @tuopuu)?

tuopuu:
@kemattil, the same question should be pointed to the other (abstract) CUDS containers as well, although they only have the 'name' attribute defined.

kemattil:
Yes, I agree.

nathanfranklin:
> None of the lattice properties (size, origin, data, etc.) are properties of the ABCLattice.
That is, should they be defined as abstract properties in ABCLattice

Currently, the attributes of all the abstract CUDS classes are listed in the docstring.  We can update them all to abstractproperty if desired. 

@tuopuu could you add `data`also to the ABCLattice docstring

tuopuu:
> could you add data also to the ABCLattice docstring

@nathanfranklin, good point. I'll add it.

kemattil:
@nathanfranklin , ok: perhaps the Dev-team can decide on this.

simphony/io/tests/test_h5_lattice.py
```diff
@@ -73,9 +108,10 @@ def cleanup(self):
         shutil.rmtree(self.temp_dir)
 
     def container_factory(self, name, type_, base_vect, size, origin):
-        return H5Lattice.create_new(
-            self.handle.root, 'foo', type_, base_vect,
-            size, origin, record=CustomRecord)
+        self.group = self.handle.create_group(
```
itziakos: Is there a reason for adding this change?

simphony/io/tests/test_h5_lattice.py
```diff
@@ -38,8 +39,10 @@ def cleanup(self):
         shutil.rmtree(self.temp_dir)
 
     def container_factory(self, name, type_, base_vect, size, origin):
+        self.group = self.handle.create_group(
```
itziakos: For the tests it does not look that there is a reason to create another group node.

itziakos:
The PR looks ok, just a few minor comments

simphony/io/tests/test_h5_lattice.py
tuopuu: In-file lattice is now created in the group pointed by self.group (which in this case is /root/foo) instead of /root/. This change can be removed from the tests, it's not relevant otherwise than moving the tables in a subgroup.

itziakos:
> None of the lattice properties (size, origin, data, etc.) are properties of the ABCLattice.
That is, should they be defined as abstract properties in ABCLattice (@nathanfranklin , @itziakos, @tuopuu)?

The short answer is yes, but I think the current work in SSB will formalize the set of properties that should be part of the API.

simphony/io/tests/test_h5_lattice.py
itziakos: Yes, it does not provide anything additional for the tests since the `handle.root` is already a group. On the contrary, using `handle.root`, makes sure that the `H5Lattice` proxy class does not depended on specific structures outside of the provided group node since any attempt to access a node before the `handle.root` with fail.

tuopuu: Actually, this is a relevant change. The in-file lattice 'name' parameter is now saved as the name of the subgroup where the lattice is stored. This is analogous how it's done with in-file particles. If the lattice is stored in root, as it was before in the tests, I don't know if the name can be set properly in that case. Also, 'container_factory' does not have a group parameter in it which would point to a coorrect subgroup. Therefore, subgroup has to be created in container_factory, before calling H5Lattice constructor.

itziakos: The `handle.root` is a group instance so it should properly support the `_v_name` parameter.

tuopuu: If handle.root is used as the group where tables are created, then to set the name properly, container_factory has to set the name of the root group to 'name', because H5Lattice doesn't have the name parameter anymore. Do you think it's better to change the name, than create a subgroup?

itziakos: hmm..., it looks that the test is a little problematic. In theory the H5Lattice should not care about the name that the parent group has, but this means that we need to revisit the test template which I think is out of the scope of this PR. lets keep the current implementation and open an issue to revisit the tests (there is always space for cleanup).

tuopuu: That's fine by me. Also, adding the data-attribute testing to the testing template would be great.

simphony/io/h5_cuds.py
```diff
@@ -158,9 +158,10 @@ def add_lattice(self, lattice):
...
                 'Lattice \'{n}\` already exists'.format(n=lattice.name))
 
-        # Create a h5lattice with all CUBA-keys defined
-        h5lat = H5Lattice.create_new(self._root.lattice, lattice.name,
-                                     lattice.type, lattice.base_vect,
+        group = self._handle.create_group('/lattice/', lattice.name)
+
+        # Create a h5_lattice with all CUBA-keys defined
+        h5lat = H5Lattice.create_new(group, lattice.type, lattice.base_vect,
                                      lattice.size, lattice.origin)
```
nathanfranklin: looks like the `data` from the added lattice is not given to the h5-lattice.

simphony/io/tests/test_h5_lattice.py
nathanfranklin: what about making the following changes to avoid `self.group`. 
one could use H5Cuds:
```self.handle = H5CUDS.open(self.filename)```
and then, change  `container_factor` to
```
    def container_factory(self, name, type_, base_vect, size, origin):
        self.name = name
        return self.handle.add_lattice(Lattice(name,
            type_, base_vect, size, origin))
```
`self.group` is removed and the spot where self.group was used could be updated to
```
    def test_initialization_from_existing_lattice_in_file(self):
        """ Checks that H5Lattice constructed from Lattice with a
        CustomRecord column description has correct attributes
        """
        lattice = self.handle.get_lattice(self.name)
        self.assertEqual(lattice.name, 'foo')
        ...
```

itziakos: I would like to avoid using H5CUDS in order to keep the testing isolated and independent of the internals of H5CUDS. This mainly the goal of unit-testing vs integration testing.

simphony/io/h5_cuds.py
tuopuu: That's true, thank you. I'll add that and related tests.

nathanfranklin:
Just left a minor comment.  Otherwise :+1:

tuopuu:
Simplified the line pointed by @nathanfranklin. I'll merge when Travis passes.

tuopuu:
Merged #120

Issue #98, opened 2015-03-13
Incosistent representation of UID


roigcarlo:
The uid representation while iterating using iter_points is different from the one returned by iteration over iter_faces().uid in FileMesh

The correct represnetation is an UUID

PR #125, opened 2015-03-27
Fixed error with uid

roigcarlo:
This PR fixes https://github.com/simphony/simphony-common/issues/98

### DIFF
--- a/simphony/io/tests/test_h5_mesh.py
+++ b/simphony/io/tests/test_h5_mesh.py
@@ -350,14 +350,8 @@ class TestH5Mesh(unittest.TestCase):
 
         self.assertItemsEqual(iedges_id, euids)
 
-        mesh_points = [p.uid for p in self.mesh.iter_points()]
-        edge_points = []
-
-        for edge in self.mesh.iter_edges():
-            for p in edge.points:
-                edge_points.append(p)
-
-        self.assertItemsEqual(set(mesh_points), set(edge_points))
+        for index, edge in enumerate(self.mesh.iter_edges(euids)):
+            self.assertItemsEqual(edge.points, edges[index].points)
 
     def test_get_all_faces_iterator(self):
         """ Checks the face iterator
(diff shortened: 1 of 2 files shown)


itziakos:
We probably need some tests to make sure that we will not have a regression of the issue

roigcarlo:
@itziakos I have extended the iterator tests to check that as is where was present. I'm not sure if this is the proper way to check it, but I am not able to find any other way to verify that the two sets of id's are the same.

 I have also added a missing test to check point_iter() without arguments.

simphony/io/tests/test_h5_mesh.py
```diff
@@ -332,6 +350,15 @@ def test_get_all_edges_iterator(self):
 
         self.assertItemsEqual(iedges_id, euids)
 
+        mesh_points = [p.uid for p in self.mesh.iter_points()]
```
itziakos: We should check that the specific uuids in each edge.points are equal to the uuids there where given at initialization. 
something like
```python
for index, edge in self.mesh.iter_edges(euids):
    self.assertItemEqual(edge.points, edges[index].points)
```

itziakos: Similar logic can be applied in the other places.

itziakos:
> @itziakos I have extended the iterator tests to check that as is where was present. I'm not sure if this is the proper way to check it, but I am not able to find any other way to verify that the two sets of id's are the same.

The check needs to be more specific.

roigcarlo:
>  something like
>  ```python
>  for index, edge in self.mesh.iter_edges(euids):
>      self.assertItemEqual(edge.points, edges[index].points)
>  ```

Actually, I was trying to avoid this type of check,

As far as I know python iterators were not designed to be indexed, so what comes into my mind to implemented this kind of check is something like:
```python
for index, edge in enumerate(self.mesh.iter_edges(euids)):
    self.assertItemsEqual(edge.points, edges[index].points)
```

Which works fine for this particular case. However, enumerating the iterators "on the fly" makes me wonder if the order in which the elements are going to be iterated and enumerated is going t be the same as the insertion order. Is it right?

itziakos:
> As far as I know python iterators were not designed to be indexed, so what comes into my mind to implement this kind of check is something like:

yes, sorry that was a stupid error, using enumerate was the intention

> Which works fine for this particular case. However, enumerating the iterators "on the fly" makes me wonder if the order in which the elements are going to be iterated and enumerated is going t be the same as the insertion order. Is it right?

Yes, so far, I have been working and implementing code under this assumption that when we provide a value of the uids argument we get the data in order. But you make a good point that we should mention that in the docstrings.

itziakos:
:+1:

Issue #122, opened 2015-03-26
"high-level" `data` is missing from H5Mesh


nathanfranklin:
This relates to #15.

PR #126, opened 2015-03-27
Implemented data in H5Mesh

roigcarlo:
This PR should implemented the missing data attribute for H5Mesh described in https://github.com/simphony/simphony-common/issues/122


nathanfranklin:
:+1:

Issue #95, opened 2015-03-11
DOC: UML diagrams of the current CUDS api


itziakos:
We should always have the current uml diagram for the CUDS objects in our documentation. Ideally the description should be in easily editable text format so that it is version controlled in our repos, it should not depend on the implementation so that it can be updated and edited independently and sphinx should be able to compile a diagram to embed it in the documentation.

To achieve these goals we can use PlantML http://plantuml.sourceforge.net/index.html which will allow us to embed in rst the uml description of the api and render them to nice uml diagrams through sphinx.

itziakos:
see examples http://plantuml.sourceforge.net/classes.html

PR #130, opened 2015-04-14
Feature uml diagrams

itziakos:
This PR add the UML diagrams for the current CUDS api description

in more detail:

- The diagrams are described using plantuml in plain text inside the source/uml folder
- The plantuml.jar file is added and a new `make` target `uml` is added to build the png files for the documentation as necessary.
- The README is updated with instraction on setting up the plantuml build enviroment
- A new page is was added for the CUDS API description using the uml diagrams
- Minor cleanup has took place to bridge differences between the abstract classes and the native implementation (mainly signature variables).


itziakos:
fixes #95

itziakos:
@roigcarlo  and @nathanfranklin can you have a look?

doc/source/cuds_definition.rst
```diff
@@ -0,0 +1,89 @@
...
+
+  **Figure. 2:** UML description of a CUDS Containers
+
+The CUDS containers, with the exception of the Lattice, are multi-type
+container (i.e. contain different types of objects). For each type the
+following operations are currently supported:
+
+* Add an item to the container
+* Get an item from the container based on the ``uid``
+* Remove an item from the container based on the ``uid``
```
roigcarlo: Actually, mesh elements / points do not support to be removed from a mesh, as this operation was not reflected in the wiki:

https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/Mesh_description
https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/Mesh_data

It shouldn't be a problem to implement but I ignore if the lack of this feature is intended or not.

itziakos: hm.. that is correct. I could think that we forgot about it. I will correct the documentation and raise an API change on the mesh workgroup

doc/source/cuds_definition.rst
```diff
@@ -0,0 +1,89 @@
...
+
+.. note::
+
+   For Lattice items do not have a ``uid`` but are accessed based on
+   ``(i,j,k)`` index tuples.
+
+.. rubric:: Snapshot principle
+
+All CUDS containers (native or proxy based) are owners of their data
+and will always return cuds components that container copy (snapshot)
```
nathanfranklin: cuds -> CUDS

"components that container a copy (snapshot)"-> "components that contain a copy (snapshot)"

doc/source/cuds_definition.rst
```diff
@@ -0,0 +1,89 @@
...
+of the internal data representation. As a result running comparing the
```
nathanfranklin: Suggestion:  The "As a result.." sentence could be changed to  "As a result, the information of an item extracted with a ``get`` operation before and and after a simulation run is not expected to be the same."

nathanfranklin:
@itziakos ,  i like the descriptions in CUDS.   looks good; i just left a few suggestions for the text.

doc/source/cuds_definition.rst
itziakos: I have updated the uml diagram of the Mesh and added more info in the exceptions note

itziakos:
@nathanfranklin and @roigcarlo  I have updated the PR, please have another look

itziakos:
One thing that is bad with this PR is that it add the plantuml.jar binary file which is not nice and problematic for versioned controlled repositories

nathanfranklin:
:+1:

itziakos:
@nathanfranklin and @roigcarlo  I have removed the planuml.jar and fixed the mesh api description. Please have another look

nathanfranklin:
:+1:

Issue #118, opened 2015-03-25
numpy is missing from requirements


nathanfranklin:
numpy package is missing from requirements (dev_requirements and README)

nathanfranklin:
closed by #132

PR #132, opened 2015-04-16
add numpy to requirements

nathanfranklin:
This addresses #118


nathanfranklin:
This PR is ready to be reviewed.

Note that I **arbitrarily** put the required version at ">=1.4.1" .  I am using 1.9.1 but see that the version of PyTables (3.1.1) that we are using lists a numpy version >= 1.4.1 (although in setup.py, they don't use a specific version).

itziakos:
:+1:

PR #134, opened 2015-04-17
Check template for mesh

itziakos:
This PR add a set template objects to check conformance for Mesh containers.

in more details:

- Separate template classes to check functionality of point, edge, face and cell operations.
- Separate template class to check Mesh attributes functionality
- Updated Mesh and H5Mesh tests to use the templates.
- Fixed bug where the number of connected points was not updated.
- Updated doc-strings to be consistent with the currently expected behaviour.
- Minor cleanups here and there.


itziakos:
@roigcarlo and @nathanfranklin this PR is now ready for review. Please note that we improved the code coverage by ~ 4%.

simphony/cuds/mesh.py
```diff
@@ -495,23 +478,12 @@ def update_point(self, point):
         KeyError
```
nathanfranklin: ValueError

itziakos: Done

nathanfranklin:
:+1:

itziakos:
@roigcarlo and @nathanfranklin, I have updated the PR to include a set of tests to target containers that do not support all the cuba keys.

nathanfranklin:
:+1:

PR #136, opened 2015-04-19
Fix incorrect commit

### LINKED ISSUE #135: Incorrect import in mesh.py
The abcmesh 

``` python
from abcmesh import *
```

should be 

``` python
from simphony.cuds.abcmesh import *
```

The former prevents the mesh class to be imported while outside the simphony module.

roigcarlo:
This fixes #135


itziakos:
:+1:

Issue #127, opened 2015-03-27
DOC: update docstrings of CUDS items iterators to reflect expected order of items


itziakos:
We should be specific on the order of the returned item from the CUDS items iterators (e.g. `iter_points()`)

For example
```

Parameters
----------------
uids : iterable or None
     When an uids is provided then points are returned in the same order the 
     uuids are returned by the iterable. If uids is None, there is no restriction on
     the order the points in the container are returned.
```

nathanfranklin:
:+1:

PR #138, opened 2015-04-23
update docstrings of CUDS items iterators to reflect expected order of items

nathanfranklin:
This addresses #127.

All the the docstrings are updated to clarify the spefic order of items by the CUDS items iterators (e.g. iter_points()).

### DIFF
--- a/simphony/cuds/abstractlattice.py
+++ b/simphony/cuds/abstractlattice.py
@@ -54,7 +54,10 @@ class ABCLattice(object):
         Parameters
         ----------
         indices : iterable set of D x int, optional
-            node index coordinates
+            When indices (i.e. node index coordinates) are provided, then nodes
+            are returned in the same order of the provided indices. If indices
+            is None, there is no restriction on the order the nodes that are
+            are returned.
 
         Returns
         -------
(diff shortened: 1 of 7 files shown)


nathanfranklin:
Notice that for mesh (ABCMesh ...), I changed the parameter names of the inter methods from "xxx_uids" to "uids".  This is already reflected in the uml diagrams

nathanfranklin:
This PR is ready to review.   @roigcarlo , @itziakos , @tuopuu , @SGGgarcia , if one of you could take a look at it before monday, that would be great.

roigcarlo:
Just a couple of comments:

If we use uid in mesh:
https://github.com/simphony/simphony-common/blob/feature_update_docstring_cuds_items_iters/simphony/cuds/mesh.py#L30
https://github.com/simphony/simphony-common/blob/feature_update_docstring_cuds_items_iters/simphony/cuds/mesh.py#L75

should be ass well "uid"

roigcarlo:
I also have notice that in the mesh class we are always referring the **uid** as **"uid"** in the docstrings,  while in the particles.py for example the **uid** is referred as **"id"**.

I think we should use either of the options but not both at the same time. It looks a little bit confusing while put it all together

nathanfranklin:
@roigcarlo , take a look.  hopefully, they are consistent

roigcarlo:
@nathanfranklin I have uploaded some more, I think it would be easier that way rather than mark them in the code.

Apart from that :+1:

nathanfranklin:
@roigcarlo Perfect!  thanks

PR #140, opened 2015-05-05
add reference to the SimPhoNy-Framework

### LINKED ISSUE #131: DOC: Reference and point to the simphony-framework
Is would be nice if the simphony-framework is also advertised in the Simphony documentation

itziakos:
This Fixes #131


roigcarlo:
:+1:

PR #141, opened 2015-05-05
Support custom cuba keys set in particle check templates

itziakos:
This PR streamlines the particle check template to support testing containers that only accept a limited
set of cuba keys.


itziakos:
@nathanfranklin this is ready for review

simphony/testing/utils.py
```diff
@@ -86,12 +93,18 @@ def create_points():
...
 
 
-def create_bonds(n=5, restrict=None):
+def create_bonds(n=5, restrict=None, particles=None):
     bond_list = []
     for i in xrange(n):
         data = create_data_container(restrict=restrict)
-        data[CUBA.VELOCITY] = i
-        ids = [uuid.uuid4() for x in xrange(n)]
+        # FIXME: This is not right since CUBA.VELOCITY might not be part of the
```
nathanfranklin: could the VELOCITY be removed now?

nathanfranklin:
Added one question about the TODO regarding VELOCITY.  Otherwise, :+1:

simphony/testing/utils.py
itziakos: We need to have a way to differentiate the data containers for the others and at the same time adapt to the fact that the CUBA key that we use should be one of the supported cuba, preferably it should be a predictable value. I have some ideas but I need a little more time to think about it. 

Let me know if you want to give it a go!

itziakos:
@nathanfranklin, I removed the need for the CUBA.VELOCITY. This PR is now ready to review

simphony/testing/utils.py
```diff
@@ -107,34 +114,53 @@ def create_data_container(restrict=None):
...
     -------
     data : DataContainer
 
 
-    """
 
+    """
     if restrict is None:
         restrict = CUBA
     data = {cuba: dummy_cuba_value(cuba) for cuba in restrict}
```
nathanfranklin: `constant` is missing from `dummy_cuba_value`

itziakos: oops, good catch

itziakos: Done

nathanfranklin:
:+1:

PR #147, opened 2015-05-14
PR: Lattice attributes are always 3D.

### LINKED ISSUE #80: CUDS: Lattice size, origin and base_vectors should be always in 3D
This will allow defining lattices that are in the XY, YZ, XZ planes. Such lattices will be also easier to visualized next to full 3D lattice instances.

tuopuu:
This PR fixed #80. Please, note that documentation is not modified yet.

Changes:
abstractlattice.py:
* replaced "D x " notation with [3] in docstrings

lattice.py:
* replaced "D x " notation with [3] in docstrings
* factory functions always create 3D Lattices. 2D lattice is embedded on the XY-plane in 3D. This is due to decision by the SSB that the 3rd coordinate is ignored for 2D lattices.
* factory functions still require correct dimensions for lattice parameters, e.g. two dimensions for the size of 2D lattice.

test_lattice.py:
* changed all 2D variables to 3D except in factory functions by adding 0 or 1

h5lattice.py:
* replaced "D x " notation with [3] in docstrings

abc_check_lattice.py:
* tests for 2D-lattices now use 3D attributes

### DIFF
--- a/simphony/cuds/lattice.py
+++ b/simphony/cuds/lattice.py
@@ -9,8 +9,8 @@ class LatticeNode(object):
 
     Attributes
     ----------
-    index : tuple of D x int
-        D dimensional node index coordinate
+    index : tuple of int[3]
+        node index coordinate
     data : DataContainer
 
     """
@@ -33,12 +33,12 @@ class Lattice(ABCLattice):
         name of the lattice
     type : str
         Bravais lattice type (should agree with the base_vect below).
-    base_vect : D x float
-        defines a Bravais lattice of dimension D = 2,3
+    base_vect : float[3]
+        defines a Bravais lattice
         (an alternative for primitive vectors).
-    size : tuple of D x int
+    size : tuple of int[3]
         number of lattice nodes in the direction of each axis.
-    origin : D x float
+    origin : float[3]
         origin of lattice
 
     """
@@ -57,7 +57,8 @@ class Lattice(ABCLattice):
 
         Parameters
         ----------
-        index : tuple of D x int (node index coordinate)
+        index : int[3]
+            node index coordinate
 
         Returns
         -------
@@ -88,11 +89,11 @@ class Lattice(ABCLattice):
 
         Parameters
         ----------
-        indices : iterable set of D x int, optional
+        indices : iterable set of int[3], optional
             When indices (i.e. node index coordinates) are provided, then nodes
             are returned in the same order of the provided indices. If indices
             is None, there is no restriction on the order the nodes that are
-            are returned.
+            returned.
 
         Returns
         -------
@@ -111,11 +112,12 @@ class Lattice(ABCLattice):
 
         Parameters
         ----------
-        index : D x int (node index coordinate)
+        index : int[3]
+            node index coordinate
 
         Returns
         -------
-        D x float
+        float[3]
 
         """
         if self._type == 'Hexagonal':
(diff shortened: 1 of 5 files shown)


tuopuu:
@itziakos, @roigcarlo, @kemattil here's the initial push to fix #80. I think there are several ways to implement the SSB decision. This is the one I thought was the simplest. Please, take a look.

There are no try/exceptions written that test the dimensionality of input parameters, but I think they could be added to the factory functions, at least. Please, let me know if you think that's a good idea to do.

itziakos:
> There are no try/exceptions written that test the dimensionality of input parameters, but I think they could be added to the factory functions, at least. Please, let me know if you think that's a good idea to do.

I am in favor of not overdoing it also.

simphony/cuds/lattice.py
```diff
@@ -149,17 +151,18 @@ def data(self, value):
...
     h : float
-        lattice spacing.
-    size : 2 x int
-        number of lattice nodes (in each axis direction).
-    origin : 2 x float (default value = (0,0))
-        lattice origin.
+        lattice spacing
+    size : int[2]
+        Number of lattice nodes in each axis direction.
+    origin : float[2], default value = (0, 0)
```
itziakos: I think origin should be x, y, z

simphony/cuds/lattice.py
```diff
@@ -149,17 +151,18 @@ def data(self, value):
 
 
 def make_hexagonal_lattice(name, h, size, origin=(0, 0)):
-    """Create and return a 2D hexagonal lattice.
+    """Create and return a 2D hexagonal lattice embedded on the XY-plane
```
itziakos: So we do not allow a XZ or YZ or generic planes just XY correct?

simphony/cuds/lattice.py
```diff
@@ -167,21 +170,23 @@ def make_hexagonal_lattice(name, h, size, origin=(0, 0)):
...
     h : float
-        lattice spacing.
-    size : 2 x int
-        number of lattice nodes (in each axis direction).
-    origin : 2 x float (default value = (0,0))
-        lattice origin.
+        lattice spacing
+    size : int[2]
+        Number of lattice nodes in each axis direction.
+    origin : float[2], default value = (0, 0)
```
itziakos: See https://github.com/simphony/simphony-common/pull/147/files#r30311721

simphony/cuds/lattice.py
```diff
@@ -189,21 +194,23 @@ def make_square_lattice(name, h, size, origin=(0, 0)):
...
-        lattice spacings (in each axis direction).
-    size : 2 x int
-        number of lattice nodes (in each axis direction).
-    origin : 2 x float (default value = (0,0))
-        lattice origin.
+    hs : float[2]
+        lattice spacings in each axis direction
+    size : int[2]
+        Number of lattice nodes in each axis direction.
+    origin : float[2], default value = (0, 0)
```
itziakos: see https://github.com/simphony/simphony-common/pull/147/files#r30311721

simphony/cuds/lattice.py
tuopuu: As I understood the SSB statement, we only support "2D lattice" on an XY-plane. Personally, I would have allowed all planar configurations, but it's simpler this way anyway. Also, one can always make a 3D lattice with size dimension 1 to any direction to create fully working planar 2D lattice.

simphony/cuds/lattice.py
tuopuu: You're right. Docstrings already say that the lattice will be embedded in 3D so the origin must be in 3D.

Issue #148, opened 2015-05-18
Duplicate hdf5 groups of the CUDS containers


itziakos:
reported by @roigcarlo 

In general, each time that a container is generated here ( for the mesh case for example ):
https://github.com/simphony/simphony-common/blob/master/simphony/io/h5_cuds.py#L211

Specifically, this error is being triggered by

test_add_get_mesh
test_mesh_rename
test_iter_mesh
test_delete_mesh
So I assume that this is happening because the python get_xxx() functions of h5_cuds.py are returning the container by copy. This does not cause problems if the original objects is "normal" object, since the group does not exist, but when python group = self._root.mesh._f_get_child(name) is not empty it duplicates the groups.

I have been able to duplicate this as well with the h5_lattice class, and it looks that h5_particles is going to behave the same way

itziakos:
@roigcarlo, I run the `test_iter_mesh` but I cannot see the duplicate hdf5 groups
```
c:\users\itziakos\appdata\local\temp\tmpgtjpxa\test.cuds (File) u''
Last modif.: 'Fri Jun 19 13:18:58 2015'
Object Tree: 
/ (RootGroup) u''
/lattice (Group) 'lattice'
/mesh (Group) 'mesh'
/particle (Group) 'particle'
/mesh/test_0 (Group) ''
/mesh/test_0/cells (Table(0,)) ''
/mesh/test_0/data (Table(1,)) ''
/mesh/test_0/edges (Table(0,)) ''
/mesh/test_0/faces (Table(0,)) ''
/mesh/test_0/item_data (Table(0,)) ''
/mesh/test_0/points (Table(0,)) ''
/mesh/test_1 (Group) ''
/mesh/test_1/cells (Table(0,)) ''
/mesh/test_1/data (Table(1,)) ''
/mesh/test_1/edges (Table(0,)) ''
/mesh/test_1/faces (Table(0,)) ''
/mesh/test_1/item_data (Table(0,)) ''
/mesh/test_1/points (Table(0,)) ''
/mesh/test_2 (Group) ''
/mesh/test_2/cells (Table(0,)) ''
/mesh/test_2/data (Table(1,)) ''
/mesh/test_2/edges (Table(0,)) ''
/mesh/test_2/faces (Table(0,)) ''
/mesh/test_2/item_data (Table(0,)) ''
/mesh/test_2/points (Table(0,)) ''
/mesh/test_3 (Group) ''
/mesh/test_3/cells (Table(0,)) ''
[...]

PR #149, opened 2015-05-20
Fix h5mesh data attribute conflict

### LINKED ISSUE #144: H5Mesh conflict between cuba data tables
It looks that the hdf5 backed mesh defines the two tables to hold the container data attirbute and the items data content on a table node with the same name. See lines:

https://github.com/simphony/simphony-common/blob/master/simphony/io/h5_mesh.py#L128
https://github.com/simphony/simphony-common/blob/master/simphony/io/h5_mesh.py#L791

The problem was not detected because the tests do not really check for that sort of interaction.

itziakos:
this PR fixes #144


simphony/io/tests/test_h5_mesh.py
```diff
@@ -106,5 +106,53 @@ def container_factory(self, name):
...
+        # when
+        self.assertEqual(len(group._f_list_nodes()), 6)
+        self.assertIsInstance(group.cells, tables.Table)
+        self.assertIsInstance(group.points, tables.Table)
+        self.assertIsInstance(group.edges, tables.Table)
+        self.assertIsInstance(group.faces, tables.Table)
+        self.assertIsInstance(group.data, tables.Table)
+        self.assertIsInstance(group.item_data, tables.Table)
+
+    def test_mesh_layout_with_new_proxy(self):
```
roigcarlo: I am not sure what is the purpose of this test.
Is this planned to check the duplicated tables of https://github.com/simphony/simphony-common/issues/148?

simphony/io/tests/test_h5_mesh.py
```diff
@@ -106,5 +106,53 @@ def container_factory(self, name):
...
+        # then
```
roigcarlo: Given

roigcarlo:
Just a couple of comments, but as soon as it passes Travis :+1:

simphony/io/tests/test_h5_mesh.py
itziakos: It is me (going and little overboard) checking that we do not create multiple nodes when we create multiple proxies of a 
already existing group. It is not directed for #148

nathanfranklin:
Just one comment: the hdf5-psuedo-uml diagrams have to be updated with the new name ("item_data")

simphony/io/tests/test_h5_mesh.py
roigcarlo: Ok nice, thx

nathanfranklin:
:+1:

PR #165, opened 2015-06-16
PyTables relared unittest are now executed only if pytables is installed

### LINKED ISSUE #143: hdf5 requirement
I blv that hdf5 should be listed as  a requirement for simphony-common, at least to pass the unit tests
e.g. on a rhel machine without hdf5, 
python2.7 -m unittest discover 
hits 8 unittest errors all of the form 
ImportError: libhdf5.so.9: cannot open shared object file: No such file or directory

roigcarlo:
This PR fixes #143.
It should prevent PyTables related cases to run if PyTables are missing


simphony/io/tests/test_data_container_table.py
```diff
@@ -1,15 +1,15 @@
...
 from simphony.io.data_container_description import Record
 from simphony.io.data_container_table import DataContainerTable
 from simphony.io.tests.abc_data_container_table_check import (
     ABCDataContainerTableCheck)
 
+try:
+    import tables
 
-class CustomData(tables.IsDescription):
+    class CustomData(tables.IsDescription):
```
roigcarlo: @itziakos I was not sure how to handle this. These classes need to be disabled somehow since they depend directly from tables,  but ```@unittest.skipIf``` cannot be used here.

I ignore if there is an elegant solution to this problem out there

itziakos: let me make a suggestion.

itziakos: I think that the best solution is to setup a `load_tests` function in `__init__.py` of the `tests` directory and use

```python
python -m unittest discover -p tests
```

see http://stackoverflow.com/questions/12487532/how-do-i-skip-a-whole-python-unittest-module-at-run-time

itziakos: This change will mean that the test module do not have to change which I think is probably the cleanest solution

itziakos: I will make a separate PR to fix the issue

itziakos:
@roigcarlo I have implemented an alternative #166 which I think is better than changing the test modules

PR #166, opened 2015-06-16
Fix test issue 143

itziakos:
This PR using an alternative way to ignore the IO tests when Pytables is not installed.

@roigcarlo, please have a look


roigcarlo:
@itziakos It looks far better than #165 thx. :+1:

PR #168, opened 2015-06-18
Make sure that there are points in the container before testing elements

itziakos:
This PR reworks the mesh template tests for element operations to work on a container where points with valid uids have been added.

### DIFF
--- a/simphony/testing/abc_check_mesh.py
+++ b/simphony/testing/abc_check_mesh.py
@@ -1,5 +1,6 @@
 import abc
 import uuid
+import random
 from functools import partial
 
 from simphony.testing.utils import (
@@ -356,6 +357,19 @@ class MeshPointOperationsCheck(MeshItemOperationsCheck):
 
 class MeshElementOperationsCheck(MeshItemOperationsCheck):
 
+    def setUp(self):
+        self.points = []
+        for multiplier in self.point_groups:
+            points = create_points()
+            for point in points:
+                point.coordinates = [
+                    value * multiplier for value in point.coordinates]
+            self.points += points
+        self.uids = [uuid.uuid4() for _ in self.points]
+        for uid, point in zip(self.uids, self.points):
+            point.uid = uid
+        MeshItemOperationsCheck.setUp(self)
+
     operation_mapping = {
         'get item': 'none',
         'add item': 'none',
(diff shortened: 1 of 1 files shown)


itziakos:
@roigcarlo, this is a minor change that should not effect testing of wrappers. However, it is a intermediate step to simplify testing until we decide on enforcing a specific behaviour https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/Operations_on_Bonds_and_other_connectivity_objects.

itziakos:
is this ok to merge?

simphony/testing/abc_check_mesh.py
```diff
@@ -384,11 +400,14 @@ def test_update_item_points(self):
         container = self.container
         uids = self._add_items(container)
         item = self.get_operation(container, uids[2])
+        point_uids = [
+            container.add_point(Point((1.0 * i, 1.0 * i, 1.0 * i)))
```
nathanfranklin: I just took a quick look at this PR and think I might be lost on its intent.  I see here that the points are added to the container here for ```test_update_item_points```,  but from the PR description I would have thought this would occur somewhere else (like MeshElementOperationsCheck's setup) so that other tests related to edges, cells and faces would be using a container that had points.  

besides this, lgtm

itziakos: The points added at the setup are used in all the tests which relay on a specific number of points available. However, the points added here make sense only for the this test. It was easier to add a bunch of points here that are going to be used and update the `points` attribute of a Mesh element.

nathanfranklin: > The points added at the setup are used in all the tests which in general relay on a specific number of points available.

I might still be missing something or incorrectly looking at things, but I don't see where the points are added to the mesh-container though.  Besides ```test_update_item_points```, the other tests (e.g. test_adding_and_getting_items for MeshEdgeOperationsCheck)  occur with a container without any points.

itziakos: Hmm...., let me have another look.

itziakos: @nathanfranklin that was a really good catch. I could not see the problem because, the related code was in my mayavi branch. It should be ok now

nathanfranklin: cool!

PR #176, opened 2015-07-06
Batch operations for particles

SGGgarcia:
Hi,
this PR covers:
- [x] Updated abstract particle class 
- [x] Updated Particles concrete class
- [x] Updated testing templates for Particles class
- [x] Updated documentation
- [x] Updated the HDF5 clases
- [x] Updated the HDF5 tests
- [x] Updated benchmark files
- [x] Added New tests with multiple ítems
- [x] Updated H5 doc

I'm making this PR so we can focus later in the HDF5 implementation, which is a bit more tricky.

### DIFF
--- a/simphony/cuds/abstractparticles.py
+++ b/simphony/cuds/abstractparticles.py
@@ -40,7 +40,8 @@ class ABCParticles(object):
         Raises
         ------
         ValueError :
-            when there is a particle with an uids that already exists in the container.
+            when there is a particle with an uids that already exists
+            in the container.
 
         Examples
         --------
@@ -75,7 +76,8 @@ class ABCParticles(object):
         Raises
         ------
         ValueError :
-            when there is a bond with an uuid that already exists in the container.
+            when there is a bond with an uuid that already exists
+            in the container.
 
         Examples
         --------
@@ -92,8 +94,8 @@ class ABCParticles(object):
         """Updates a set of particles from the provided iterable.
 
         Takes the uids of the particles and searches inside the container for
-        those particles. If the particles exists, they are replaced in the container.
-        If any particle doesn't exist, it will raise an exception.
+        those particles. If the particles exists, they are replaced in the
+        container. If any particle doesn't exist, it will raise an exception.
 
         Parameters
         ----------
@@ -108,7 +110,7 @@ class ABCParticles(object):
 
         Examples
         --------
-        Given a set of Particle objects that already exists in the Particles container
+        Given a set of Particle objects that already exists in the container
         (taken with the 'get_particle' method for example), just call the
         function passing the Particle items as parameter.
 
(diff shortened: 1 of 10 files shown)


nathanfranklin:
For the documentation, the uml diagrams need to be updated as well.

itziakos:
@SGGgarcia, please have a look at the discussion in https://github.com/simphony/simphony-common/pull/178. If people agree we can also convert the `get_xxx` -> `get_xxxs` using the proposed pattern. It is also ok if such change happens at a later PR.

simphony/cuds/abstractparticles.py
```diff
@@ -16,140 +16,143 @@ class ABCParticles(object):
...
 
         >>> particles = Particles(name="foo")
         >>> ...
-        >>> bond = particles.get_bond(uid)
-        >>> ... #do whatever you want with the bond
-        >>> particles.update_bond(bond)
+        >>> bond1 = particles.get_bond(uid1)
+        >>> bond2 = particles.get_bond(uid2)
+        >>> ... #do whatever you want with the bonds
+        >>> particles.update_bond([bond1, bond2])
```
itziakos: this should be `update_bonds`

simphony/testing/abc_check_particles.py
```diff
@@ -393,7 +389,7 @@ def test_exeception_when_updating_bond_with_incorrect_id(self):
         # then
         with self.assertRaises(ValueError):
             # when
-            self.container.update_bond(bond)
+            self.container.update_bonds([bond])
 
     def test_remove_bond(self):
```
itziakos: I think that we should also rename the test method names to reflect the new method names

itziakos:
Please also add tests for the new methods using more than one items. We do not have to replace the old, just also add some new ones.

nathanfranklin:
`cuds_file_bench.py` also needs to be updated.

SGGgarcia:
@itziakos @nathanfranklin I think everything is done, tell me otherwise :+1:

simphony/testing/utils.py
```diff
@@ -80,6 +80,18 @@ def create_particles(n=10, restrict=None):
     return particle_list
 
 
+def create_particles_with_id(n=10, restrict=None):
```
itziakos: We need to add tests for these functions.

itziakos:
Just one comment, apart from that it looks ok.

simphony/testing/utils.py
SGGgarcia: Since I think we will have to add them to the new `testing/test/test_utils.py` file that is not part of the master branch yet, can we add them after we merge #181 ?

itziakos: Good point, please add a separate issue so that we can track the remaining work.

SGGgarcia: Done in #184

Issue #164, opened 2015-06-16
Provide method to return the count of contained items in the CUDS containers


itziakos:
A set of CUBA keywords will be added for the low level CUDS items.

```python
POINT
PARTICLE
BOND
EDGE
FACE
CELL
NODE
```

New method in CUDS containers

```python
  def count_of(self, item_type):
      """ Return the count of item_type in the container.

      Parameter
      ---------
      item_type : CUBA
          The CUBA enum of the type of the items to return the count of.         

      Returns
      -------
      count : int
          The number of items of item_type in the container.

      Raises
      ------
      ValueError :
          If the type of the item is not supported in the current container. 
      
      """
```

**tasks**:

- [ ] update the KEYWords yml and regenerate the CUBA enum
- [ ] update the uml diagrams
- [ ] Augment test template
- [ ] update the abstract classes
- [ ] update the concrete classes
- [ ] update the hdf5 back classes

PR #177, opened 2015-07-06
Add count operations

roigcarlo:
This PR implements #164

- [x] update the KEYWords yml and regenerate the CUBA enum
- [x] Augment test template
- [x] update the abstract classes
- [x] update the concrete classes
- [x] update the hdf5 back classes


simphony/core/keywords.py
```diff
@@ -554,4 +554,60 @@
...
+        name='Cell',
+        description='Cell item identifier',  # noqa
+        domain=['ATM', 'DEM', 'FEM', 'FVM', 'LBM', 'SPH', 'VIS'],
+        key='CELL',
+        number=73,
+        shape=[1],
+        dtype=numpy.int32),
+    'NODE': Keyword(
+        name='Node',
+        description='Nod item identifier',  # noqa
```
nathanfranklin: Nod -> Node

simphony/testing/abc_check_particles.py
```diff
@@ -229,6 +230,26 @@ def test_exception_on_iter_particles_when_passing_wrong_ids(self):
         # then
         self.assertEqual(particle.uid, self.particle_list[-1].uid)
 
+    def test_count_of_particles(self):
+        # given
+        container = self.container
+        particle_list = self.particle_list
+
+        # then
+        # The order of iteration is not important in this case.
```
nathanfranklin: delete old comment

simphony/testing/abc_check_particles.py
```diff
@@ -229,6 +230,26 @@ def test_exception_on_iter_particles_when_passing_wrong_ids(self):
...
+        count_original = len(particle_list)
+        count_container = container.count_of(CUBA.PARTICLE)
+        self.assertEqual(count_original, count_container)
+
+    def test_count_of_particles_passing_unsupported_type(self):
+        # given
+        container = self.container
+
+        # then
+        # The order of iteration is not important in this case.
```
nathanfranklin: delete old comment

simphony/testing/abc_check_particles.py
```diff
@@ -457,3 +478,23 @@ def test_exception_on_iter_bonds_when_passing_wrong_ids(self):
...
         # then
         self.assertEqual(bond.uid, self.bond_list[-1].uid)
+
+    def test_count_of_bonds(self):
+        # given
+        container = self.container
+        bond_list = self.bond_list
+
+        # then
+        # The order of iteration is not important in this case.
```
nathanfranklin: delete old comment

simphony/testing/abc_check_particles.py
```diff
@@ -457,3 +478,23 @@ def test_exception_on_iter_bonds_when_passing_wrong_ids(self):
...
+        count_original = len(bond_list)
+        count_container = container.count_of(CUBA.BOND)
+        self.assertEqual(count_original, count_container)
+
+    def test_count_of_bonds_passing_unsupported_type(self):
+        # given
+        container = self.container
+
+        # then
+        # The order of iteration is not important in this case.
```
nathanfranklin: delete old comment

simphony/cuds/lattice.py
```diff
@@ -132,6 +137,34 @@ def get_coordinate(self, index):
...
+        ------
+        ValueError :
+            If the type of the item is not supported in the current
+            container.
+
+        """
+
+        if item_type in self._allowed_item_types:
+            if item_type == CUBA.NODE:
+                return reduce(lambda x, y: x*y, self._size)
```
tuopuu: @roigcarlo, you could replace this by
´´´python
    return np.prod(self._size)
´´´
because numpy is already imported.

simphony/core/cuba.yml
```diff
@@ -474,3 +474,52 @@
   number: 67
   shape: [1]
   type: double
+- description: Point item identifier
+  domain: [ATM, DEM, FEM, FVM, LBM, SPH, VIS]
```
itziakos: Please restrict the domain for the new attributes. If I am correct they are not valid for all.

simphony/cuds/abstractmesh.py
```diff
@@ -15,76 +15,386 @@ class ABCMesh(object):
...
+        ------
+        ValueError :
+            If the point was not found in the mesh
+
+        """
 
     @abstractmethod
     def update_edge(self, edge):
-        pass
+        """ Updates the information of a edge.
```
itziakos: `a edge` -> `an Edge`

simphony/io/data_container_description.py
```diff
@@ -70,16 +70,23 @@ class Data(tables.IsDescription):
     ln_of_restitution_coefficient = tables.Float64Col(pos=63)
     rolling_friction = tables.Float64Col(pos=64)
     volume_fraction = tables.Float64Col(pos=65)
+    point = tables.Int32Col(pos=66)
```
itziakos: We should probably not include these to the table

simphony/io/h5_lattice.py
```diff
@@ -124,6 +129,34 @@ def iter_nodes(self, indices=None):
...
+        Raises
+        ------
+        ValueError :
+            If the type of the item is not supported in the current
+            container.
+
+        """
+
+        if item_type in self._allowed_item_types:
+            if item_type == CUBA.NODE:
```
itziakos: We probably do not need this `if block here`.

simphony/io/h5_mesh.py
```diff
@@ -139,6 +140,13 @@ def __init__(self, group, meshFile):
         if "cells" not in self._group:
             self._create_cells_table()
 
+        self._allowed_item_types = [
```
itziakos: how about 
```python
self._items_count = {
     CUBA.POINT : lambda  : self._group.point.rows,
      ....
}
```     

then you can do

```
try:
    return self_item_count[item_type]()
except KeyError:
    raise ...

Comment (itziakos) [inline simphony/io/h5_mesh.py]: You can probably repeat the same pattern for the other Containers.

**Comment by itziakos** [comment]
@roigcarlo, PR so far looks good, some minor suggestions.

File: simphony/io/data_container_description.py (hunk 9 above)
Comment (nathanfranklin) [inline simphony/io/data_container_description.py]: I would think that we should include these in the table for the time-being.  

@itziakos, i think you proposed having a separate enum to describe this counting information at the m18 meeting.  i think that would be a good idea and would lead to them being removed from being valid keys of the DataContainer.

Comment (itziakos) [inline simphony/io/data_container_description.py]: Sorry, a forgot about that. Yes, you make a very good point. How about calling this new `enum` as `CUBAItems `

File: simphony/core/cuba.yml (hunk 7 above)
Comment (roigcarlo) [inline simphony/core/cuba.yml]: Agree, but I am not aware of exactly which types are correct for each domain.
I can figure out some of them by looking at the wrappers but I'm not sure this is the best way to proceed. 

It would be nice if the information is updated in the wiki: 

https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/API_proposal:_CUDS_containers_and_number_of_items 

or 

https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/API_proposal:_CUDS_containers_and_number_of_items

File: simphony/io/data_container_description.py (hunk 9 above)
Comment (nathanfranklin) [inline simphony/io/data_container_description.py]: `CUDSItem` :+1:

File: simphony/cuds/mesh.py [hunk 12]
```diff
@@ -676,6 +684,33 @@ def has_cells(self):
...
+        ------
+        ValueError :
+            If the type of the item is not supported in the current
+            container.
+
+        """
+
+        try:
+            return len(self._items_count[item_type]())
+        except:
```
itziakos: Please use specific exception types `except KeyError`

roigcarlo: Maybe we should go back to:

```python
if item_type in self._items_count:
    ....
else:
    raise ValueError(msg)
```

It stills remove the double ´´´if´´´ and also,  if I explicitly capture the KeyError exceptions with the actual code I would have t re-raise them with a different message. We are not doing this anywhere else through the code.

itziakos: I think there is a confusion. My comment is about the bare `try.. except:` which should be always avoided unless we really want to ignore all and any type of errors which is not the case.

What I am proposing is:

```python
try:
      return len(self._items_count[item_type]())
except KeyError:
      raise ValueError(....)
```

itziakos: >  if I explicitly capture the KeyError exceptions with the actual code. I would have t re-raise them with a different message. We are not doing this anywhere else through the code.

I am not sure why this is a problem. based on the currentl implementation the KeyError is only going to be raised if the item_type is not in self._items_count.

itziakos: > We are not doing this anywhere else through the code.

This is not true there a number of places where we specifically capture some types of errors e.g. https://github.com/simphony/simphony-common/blob/master/simphony/io/h5_lattice.py#L102

roigcarlo: > I am not sure why this is a problem

I wasn't sure if this is a common practice in python. No problem then.

> This is not true there a number of places where we specifically capture some types of errors e.g. https://github.com/simphony/simphony-common/blob/master/simphony/io/h5_lattice.py#L102

Miss that, :+1:

itziakos: In anycase if you fill that capturing KeyError is fragile (which sounds a little extreme) then please use an `if...else`. In anycase a bare `except` is not right.

itziakos: > I wasn't sure if this is a common practice in python. No problem then.

To be honest it all depends on the specifics of the code. Exception handling is always difficult and you can find arguments for both using `try..except` and defensive programming. If for example the operation in the lambda function  could return also a `KeyError` for some other reason then the `try..except` operation would not be wrong but it would be misleading.
The only thing that is always wrong is the simple `except:`.

itziakos:
This PR has some conflicts with master

itziakos:
This PR has conflicts with the current master

roigcarlo:
@itziakos Thx, i will solve the conflict as soon as I finish with the template tests for the engines.

itziakos:
@roigcarlo please merge as soon as you fix the conficts

roigcarlo:
@itziakos This should also fix problems with test_utils.py (and hence #194). If everything is ok we can merge as soon as it passes the tests.

roigcarlo:
:+1:

Issue #175, opened 2015-06-30
Provide batch operations for Particles CUDS API


itziakos:
```python
class ABCParticles(self):

    def add_particles(iterable):
        """ Add a set of particles from the provided iterable.

        """

    def update_particles(iterable):
        """ Update a set of particles from the provided iterable.

        """

    def remove_particles(iterable):
        """ Update a set of particles from the provided iterable.

        """

    def remove_particles(uids):
        """ Remove the particles with the provided uids.

        """

    def add_bonds(iterable):
        """ Add a set of bonds from the provided iterable.

        """

    def update_bonds(iterable):
        """ Update a set of bonds from the provided iterable.

        """

    def remove_bonds(uids):
        """ Remove the bonds with the provided uids.

        """
```

**notes**:

- [ ] This will replace the `add_xxxx`,  `update_xxxx` and `remove_xxxx` methods
- [ ] The Test templates need to be updated
- [ ] The Concrete Classes need to be updated
- [ ] Documentation needs to be updated

Issue #174, opened 2015-06-30
Provide batch operations for the Mesh CUDS API


itziakos:
```python

class ABCMesh(self):

    def add_points(iterable):
        """ Add a set of points from the provided iterable.

        """

    def update_points(iterable):
        """ Update a set of points from the provided iterable.

        """

    def remove_points(uids):
        """ Remove the points with the provided uids.

        """

    def add_edges(iterable):
        """ Add a set of edges from the provided iterable.

        """

    def update_edge(iterable):
        """ Update a set of edges from the provided iterable.

        """

    def remove_edges(uids):
        """ Remove the edges with the provided uids.

        """

    def add_faces(iterable):
        """ Add a set of faces from the provided iterable.

        """

    def update_faces(iterable):
        """ Update a set of faces from the provided iterable.

        """

    def remove_faces(uids):
        """ Remove the faces with the provided uids.

        """

    def add_cells(iterable):
        """ Add a set of cells from the provided iterable.

        """

    def update_cells(iterable):
        """ Update a set of cells from the provided iterable.

        """

    def remove_cells(uids):
        """ Remove the cells with the provided uids.

        """
```

**notes**:
- [ ] This will replace the `add_xxxx`, `update_xxxx` and `remove_xxxx` methods
- [ ] The Test templates need to be updated
- [ ] The Concrete Classes need to be updated
- [ ] Documentation needs to be updated

PR #178, opened 2015-07-07
PR: Batch operations for lattice.

tuopuu:
Changing update_node to update_nodes. The code, tests, docstrings and uml diagrams should be all done in this commit.


simphony/cuds/lattice.py
```diff
@@ -72,19 +72,21 @@ def get_node(self, index):
...
-    def update_node(self, lat_node):
-        """Update the corresponding lattice node (data copied).
+    def update_nodes(self, nodes):
+        """Update the corresponding lattice nodes (data copied).
 
         Parameters
         ----------
-        lat_node : reference to a LatticeNode object
-            data copied from the given node
+        nodes : an iterator to a LatticeNode objects
```
itziakos: nodes is an `iterable`

simphony/io/h5_lattice.py
```diff
@@ -86,22 +86,23 @@ def get_node(self, index):
...
 
-    def update_node(self, node):
+    def update_nodes(self, nodes):
         """ Updates H5Lattice data for a LatticeNode
 
         Parameters
         ----------
-        node : LatticeNode
-            reference to LatticeNode object
+        nodes : iterator of LatticeNodes
```
itziakos: see https://github.com/simphony/simphony-common/pull/178/files#r34018783

itziakos:
I wonder if we should also have `get_nodes`

tuopuu:
> I wonder if we should also have get_nodes

Probably yes. This of course should apply to other CUDS containers as well. We can delay this merge, if we want to add get_xxxs at the same time.

Thanks for the corrections, @itziakos. I'll re-push soon.

roigcarlo:
@itziakos @tuopuu What would be the signature of get_xxxs? Wouldn't it be the same as iter_nodes()?

SGGgarcia:
> @itziakos @tuopuu What would be the signature of get_xxxs? Wouldn't it be the same as iter_nodes()?

I was thinking the same.
Anyway, about having `get_xxxs`, if we are changing the API to work with multiple items, maybe we should also consider the `has_xxxs` ?

nathanfranklin:
Would `get_xxxs` method just return a list of the entities with the provided uids?  Besides having all methods being plural like `add_xxxs` , what would the benefit to the user be?  If the user wanted to access multiple entities wouldn't he or she just use the `iter_xxxs()` method.

itziakos:
> Would get_xxxs method just return a list of the entities with the provided uids? Besides having all methods being plural like add_xxxs , what would the benefit to the user be? If the user wanted to access multiple entities wouldn't he or she just use the iter_xxxs() method.

Yes, get_xxxs will return a list instead of an iterator. This is a small difference and in short means that 

```python
def get_items(uids):
      return [item for item in self.iter_items(uids)]
```

There are some functions that treat iterators different than lists (e.g. lists have a count). But you are correct that at this stage this change will be mainly for consistency.

itziakos:
> Anyway, about having get_xxxs, if we are changing the API to work with multiple items, maybe we should also consider the has_xxxs ?

While this is a good point, I think that it is not a common pattern to have such an operation. Yet I am not against it if it is decided.

roigcarlo:
Just a comment related with the UML images. Since they will have to be regenerated every time we make a modification in the API, and provided that at least #164 #174, #175, #176 will need to do such changes I would prefer to to generate them in a separated PR before the release with all the changes together rather than separately.
If not I think it will be kind of messy for the other branches once we begin to merge them.

tuopuu:
> generate them in a separated PR

That's OK for me. I can remove them from this PR.

itziakos:
> generate them in a separated PR

Yes, it makes sense

Issue #179, opened 2015-07-08
Update UML before release


roigcarlo:
UML images reflecting the changes in Milestone 0.2.0 should be generated before issuing a new release.

Issue #156, opened 2015-06-03
versioning of HDF5 files (H5CUDS)


nathanfranklin:
The file format should be extended with some version information and then to use that information for better error messages etc..  

see simphony/simphony-openfoam#26

itziakos:
Yes versioned hdf5 layout is something very important, I suggest that this issue get a high priority.

Please note that version should have two levels. A version for the H5CUDS layout, and a separate version for each CUDS container type. The best solution is to have this added as an cuds_version attribute in the top level group of the H5CUDS group and the CUDS container's group.

itziakos:
**Design proposal**

*Versioning*

In CUDS HDF5 layout version is provided in two levels:

1. Top level version of the main group (i.e. root) where the container types are grouped. This is the expected layout of the file as defined in H5CUDS.
2. The CUDS container hdf5 layout version separate for each CUDS container. 

The version should be stored in the top level group node as a node attribute names `cuds_version` and the value should be a single integer.

*Behaviour*

When a H5 object is asked to read from from a file or node (for CUDS containers) it will check the version number of raise if that particular version layout is not supported. 

*Backwords compativility*

To simplify the HDF5 cuds implementation and testing:
- Only a single individual version (the latest) should be supported for live (i.e. online) read/write of H5 top level objects from a file. 
- Backwards compatibility will be provided only at offline operation by separate helper conversion script that will know how to convert between versions that arrive as part of a stable release.
- Version numbers are incremented only in once in a *simphony-common* release cycle, thus intermediate changes of the HDF5 layout during the development process of a new release will only change HDF5 layout versions only once. 
- Optionally, The version number of the H5CUDS file is incremented when the main layout changes or the layout of the cuds containers change. This will simplify the implementation of the conversion script.

nathanfranklin:
Sounds good.

 > Optionally, The version number of the H5CUDS file is incremented when the main layout changes or the layout of the cuds containers change. This will simplify the implementation of the conversion scrip

I prefer the above approach to changing the version numbers.

nathanfranklin:
> In CUDS HDF5 layout version is provided in two levels:
>   Top level version of the main group (i.e. root) where the container types are grouped. This is the expected layout of the file as defined in H5CUDS.
>    The CUDS container hdf5 layout version separate for each CUDS container.

As you mentioned, this gives us an integer attribute called `cuds_version` for
 * main group (root)
 * particles
 * mesh
 * lattice

[...]

PR #180, opened 2015-07-08
Add version information to CUDS HDF5 file

nathanfranklin:
This PR addresses #156 by adding version information to the CUDS HDF5 file.  Specifically, a 'cuds_version' was added as an HDF5-attribute of type `int` to:
 *  *"1. Top level version of the main group (i.e. root) where the container types are grouped. This is the expected layout of the file as defined in H5CUDS."*
 *  *"2. The CUDS container hdf5 layout version separate for each CUDS container. "*

Only a single individual version (the latest) is supported and an exception is thrown when reading files containing a different version.

See discussions in issue #156 for more detail on versioning. In general, when a developer changes the top level layout (i.e. root), then the top level 'cuds-version' should be incremented for that release of SimPhoNy.  When a developer changes the layout of the CUDS container components, that's  'cuds_version' should be incremented for that release of SimPhony and the top-level version (i.e. root) should also be incremented.   

This PR doesn't update the UML images (see #179)

### DIFF
--- a/simphony/io/h5_cuds.py
+++ b/simphony/io/h5_cuds.py
@@ -4,7 +4,7 @@ from simphony.io.h5_particles import H5Particles
 from simphony.io.h5_mesh import H5Mesh
 from simphony.io.h5_lattice import H5Lattice
 
-_file_version = 1
+H5_FILE_VERSION = 1
 
 
 class H5CUDS(object):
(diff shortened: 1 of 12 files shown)


nathanfranklin:
@itziakos @roigcarlo @tuopuu @SGGgarcia ,  this is ready to be reviewed and merged.  thanks.  

(i am out of the office till tuesday)

simphony/io/h5_cuds.py
```diff
@@ -4,6 +4,8 @@
 from simphony.io.h5_mesh import H5Mesh
 from simphony.io.h5_lattice import H5Lattice
 
+_file_version = 1
```
itziakos: I would suggest to capitalize and probably rename to `H5FILE_VERSION`

itziakos:
one minor comment from me

SGGgarcia:
Looks ok :+1:  Same suggestion as Ioannis.

nathanfranklin:
@itziakos , @SGGgarcia :  Thanks.  Made the change.

Issue #162, opened 2015-06-16
Naming of the testing template classes is not consistent


itziakos:
We should make sure that the naming and function of the template testing classes is consistent

PR #185, opened 2015-07-13
Cleanup PR

itziakos:
This PR addresses #162 and makes naming a little more consistent for the CUDS classes

in more details:
- The abstract class filenames are converted from `abstract -> abc` to be consistent with the class names they define
- Fix a file handle leak when the H5CUDS file we try to open does not have the valid version.
- Update the names of the template classes
- Augment the tests for the Particle container.
- All template classes use an abstract method for restricting the CUBA keys that are tested.
- Remove unused code.
- Clean up some testing templates
- Update docstrings


itziakos:
This PR should be merge after the new API providing PRs. It will easier to update this PR than all the others that are currently pending

SGGgarcia:
One minor question, is one of the commits fixing one little bug about the `test_incorrect_version` test failing?

itziakos:
Yes, the test was failing because the file handle was leaked and the file was not closed when an error was raised. As a result when the test tried to cleanup the directory Windows 7 will not allow the operation to continue because of open file-handles (apparently linux does not care about open handles)

SGGgarcia:
Ok, because it was failing for me too, thanks

SGGgarcia:
Oh I was going to comment about merging this before the others to fix the tests but this doesnt affect Travis anyway, so ok :+1:

itziakos:
@SGGgarcia, @roigcarlo  and @nathanfranklin, I have updated the PR name since there are a number of cleanup commits here. Please have a look

roigcarlo:
@itziakos Just a suggestion. I'm detecting that here, and also in some others PR that have been merged these days, we have a problem with ```@abstractmethods```.

Since it seems that codecovs cannot handle that well I would suggest to raise a NotImplementedError:

```python
@abstractmethods
def foo():
    raise NotImplementedError
```

This should allow codecovs to correctly process the abstract clases. This also would help us to detect not implemented code in a future.

itziakos:
@roigcarlo, the issue is with the `patch` setup of codecov. The problem is that coverage will collect information that these methods have not been called (abstract methods are not called). 

> Since it seems that codecovs cannot handle that well I would suggest to raise a NotImplementedError:

Adding NotImplemented error will not make the issue go away. The abstract methods will still not been called. The discrepancy is because codecov does not know about the coveragerc configuration.

> This also would help us to detect not implemented code in a future.

All methods decorated with absrtactmethod will raise an exception anyway.

I think the best solution is to not use the patch option (the overall coverage check will still be there).

itziakos:
Note that the `patch coverage option` for codecov was added very recently and I activated earlier last week. It looks that there is still some work to be done from their side.

roigcarlo:
I see, thx for clarifying it, :+1:

simphony/testing/abc_check_particles.py
```diff
@@ -3,14 +3,86 @@
...
+        container = self.container
+        data = create_data_container()
+        expected_data = create_data_container(restrict=self.supported_cuba())
+
+        # when
+        container.data = data
+
+        # then
+        self.assertEqual(container.data, expected_data)
+        self.assertIsNot(container.data, expected_data)
```
nathanfranklin: this second assert is not needed, right?

itziakos: We need to check that the two DataContainers are equal but not the same object.

itziakos: oops you are right

nathanfranklin:
:+1:

SGGgarcia:
Looks ok

PR #182, opened 2015-07-13
Fix issue 159

roigcarlo:
This PR implements #159 

- [x] Update documentation on new API
- [x] Update ABC description of the ABCEngine
- [ ] Add a template for testing Engines
- [x] Update api of the H5CUDS file

PR #189, opened 2015-07-14
Simplify ABCEngine Cont.

roigcarlo:
This PR implements the rest of the #182 which was accidentaly merged.

TODO:
- [x] Extend Mesh and Particles compare tests with Bonds and Elements
- [x] Create Compare test for Lattice

### DIFF
--- a/simphony/io/tests/test_h5_cuds.py
+++ b/simphony/io/tests/test_h5_cuds.py
@@ -6,6 +6,9 @@ import tempfile
 import tables
 
 from simphony.io.h5_cuds import H5CUDS
+from simphony.io.h5_mesh import H5Mesh
+from simphony.io.h5_particles import H5Particles
+from simphony.io.h5_lattice import H5Lattice
 
 from simphony.testing.abc_check_engine import (
     ParticlesCudsCheck, MeshCudsCheck,
@@ -83,6 +86,12 @@ class TestParticlesCudsOperations(ParticlesCudsCheck, unittest.TestCase):
         filename = os.path.join(self.temp_dir, name)
         return H5CUDS.open(filename, mode)
 
+    def check_instance_of_dataset(self, ds):
+        """ Check if a dataset is instance of a class
+        """
+
+        self.assertTrue(isinstance(ds, H5Particles))
+
 
 class TestMeshCudsOperations(MeshCudsCheck, unittest.TestCase):
 
@@ -90,12 +99,24 @@ class TestMeshCudsOperations(MeshCudsCheck, unittest.TestCase):
         filename = os.path.join(self.temp_dir, name)
         return H5CUDS.open(filename, mode)
 
+    def check_instance_of_dataset(self, ds):
+        """ Check if a dataset is instance of a class
+        """
+
+        self.assertTrue(isinstance(ds, H5Mesh))
+
 
 class TestLatticeCudsOperations(LatticeCudsCheck, unittest.TestCase):
 
     def container_factory(self, name, mode='w'):
         filename = os.path.join(self.temp_dir, name)
         return H5CUDS.open(filename, mode)
+
+    def check_instance_of_dataset(self, ds):
+        """ Check if a dataset is instance of a class
+        """
+
+        self.assertTrue(isinstance(ds, H5Lattice))
 
 
 if __name__ == '__main__':
--- a/simphony/io/tests/test_h5_cuds.py
+++ b/simphony/io/tests/test_h5_cuds.py
@@ -78,15 +78,24 @@ class TestH5CUDSVersions(unittest.TestCase):
 
 
 class TestParticlesCudsOperations(ParticlesCudsCheck, unittest.TestCase):
-    pass
+
+    def container_factory(self, name, mode='w'):
+        filename = os.path.join(self.temp_dir, name)
+        return H5CUDS.open(filename, mode)
 
 
 class TestMeshCudsOperations(MeshCudsCheck, unittest.TestCase):
-    pass
+
+    def container_factory(self, name, mode='w'):
+        filename = os.path.join(self.temp_dir, name)
+        return H5CUDS.open(filename, mode)
 
 
 class TestLatticeCudsOperations(LatticeCudsCheck, unittest.TestCase):
-    pass
+
+    def container_factory(self, name, mode='w'):
+        filename = os.path.join(self.temp_dir, name)
+        return H5CUDS.open(filename, mode)
 
 
 if __name__ == '__main__':
--- a/simphony/io/tests/test_h5_cuds.py
+++ b/simphony/io/tests/test_h5_cuds.py
@@ -181,6 +181,7 @@ class TestParticlesCudsOperations(ParticlesCudsCheck, unittest.TestCase):
 
     def setUp(self):
         ParticlesCudsCheck.setUp(self)
+        self.temp_dir = tempfile.mkdtemp()
         self.engines = []
 
     def engine_factory(self):
@@ -204,6 +205,7 @@ class TestMeshCudsOperations(MeshCudsCheck, unittest.TestCase):
 
     def setUp(self):
         MeshCudsCheck.setUp(self)
+        self.temp_dir = tempfile.mkdtemp()
         self.engines = []
 
     def engine_factory(self):
@@ -227,6 +229,7 @@ class TestLatticeCudsOperations(LatticeCudsCheck, unittest.TestCase):
 
     def setUp(self):
         LatticeCudsCheck.setUp(self)
+        self.temp_dir = tempfile.mkdtemp()
         self.engines = []
 
     def engine_factory(self):
--- a/simphony/io/tests/test_h5_cuds.py
+++ b/simphony/io/tests/test_h5_cuds.py
@@ -16,8 +16,8 @@ from simphony.io.h5_particles import H5Particles
 from simphony.io.h5_lattice import H5Lattice
 
 from simphony.testing.abc_check_engine import (
-    ParticlesCudsCheck, MeshCudsCheck,
-    LatticeCudsCheck)
+    ParticlesEngineCheck, MeshEngineCheck,
+    LatticeEngineCheck)
 
 
 class TestH5CUDS(unittest.TestCase):
@@ -177,10 +177,10 @@ class TestH5CUDSVersions(unittest.TestCase):
             H5CUDS.open(self.existing_filename)
 
 
-class TestParticlesCudsOperations(ParticlesCudsCheck, unittest.TestCase):
+class TestParticlesCudsOperations(ParticlesEngineCheck, unittest.TestCase):
 
     def setUp(self):
-        ParticlesCudsCheck.setUp(self)
+        ParticlesEngineCheck.setUp(self)
         self.temp_dir = tempfile.mkdtemp()
         self.engines = []
 
@@ -201,10 +201,10 @@ class TestParticlesCudsOperations(ParticlesCudsCheck, unittest.TestCase):
             engine.close()
 
 
-class TestMeshCudsOperations(MeshCudsCheck, unittest.TestCase):
+class TestMeshCudsOperations(MeshEngineCheck, unittest.TestCase):
 
     def setUp(self):
-        MeshCudsCheck.setUp(self)
+        MeshEngineCheck.setUp(self)
         self.temp_dir = tempfile.mkdtemp()
         self.engines = []
 
@@ -225,10 +225,10 @@ class TestMeshCudsOperations(MeshCudsCheck, unittest.TestCase):
             engine.close()
 
 
-class TestLatticeCudsOperations(LatticeCudsCheck, unittest.TestCase):
+class TestLatticeCudsOperations(LatticeEngineCheck, unittest.TestCase):
 
     def setUp(self):
-        LatticeCudsCheck.setUp(self)
+        LatticeEngineCheck.setUp(self)
         self.temp_dir = tempfile.mkdtemp()
         self.engines = []
 
--- a/simphony/io/tests/test_h5_cuds.py
+++ b/simphony/io/tests/test_h5_cuds.py
@@ -1,10 +1,15 @@
 import unittest
 import os
+import uuid
 from contextlib import closing
 import shutil
 import tempfile
 import tables
 
+from simphony.cuds.particles import Particle, Particles
+from simphony.cuds.mesh import Point, Mesh
+
+from simphony.core.cuba import CUBA
 from simphony.io.h5_cuds import H5CUDS
 from simphony.io.h5_mesh import H5Mesh
 from simphony.io.h5_particles import H5Particles
@@ -20,6 +25,12 @@ class TestH5CUDS(unittest.TestCase):
     def setUp(self):
         self.temp_dir = tempfile.mkdtemp()
         self.maxDiff = None
+        self.particles = []
+        self.points = []
+        for i in xrange(10):
+            self.particles.append(
+                Particle((1.1*i, 2.2*i, 3.3*i), uid=uuid.uuid4()))
+            self.points.append(Point((1.1*i, 2.2*i, 3.3*i), uid=uuid.uuid4()))
 
     def tearDown(self):
         shutil.rmtree(self.temp_dir)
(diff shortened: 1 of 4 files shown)


roigcarlo:
@itziakos @nathanfranklin @SGGgarcia 

This PR continues the work with #182. I have followed your advice and created functions to compare Particles, Mesh and Lattices.

P.S. It looks like codecov fails but actually it has 91.72%...

simphony/testing/abc_check_engine.py
```diff
@@ -35,46 +45,48 @@ def create_dataset_items(self):
...
-    def check_is_usable_dataset(self, ds, items):
-        """ Check if a dataset is usable
-        """
+    operation_mapping = {
+        'compare datasets': 'none',
+        'add item': 'none'}
+
+    def compare_operation(self, *args, **kwrds):
+        method = getattr(
+            sys.modules[__name__],
```
itziakos: Why go to `sys.modules`?

itziakos:
please add tests for compare methods

simphony/testing/abc_check_engine.py
roigcarlo: I want to compare datasets here:

https://github.com/simphony/simphony-common/blob/873fd3378826024d3e2993836f5933cca9be0f5a/simphony/testing/abc_check_engine.py#L79

but I cannot use ```assertEqual``` as _reference_ is an instance of let's say ```Particles``` and the container retrieved _ds_ is an instance of ```H5Particles```.

So either I call ```sys.modules``` or I import all ```simphony.testing.utils``` and then use ```utils``` as the module.

roigcarlo:
@itziakos This should cover all possible tests for the new compare functions as well as some other missing tests.

Even so, I am not sure of which are the critical interactions lattices so I only added basic tests.

simphony/testing/tests/test_utils.py
```diff
@@ -356,6 +841,18 @@ def test_create_particles(self):
                 particle.data, create_data_container(constant=index),
                 testcase=self)
 
+    def test_create_particles_with_id(self):
```
SGGgarcia: @roigcarlo I have already made a small PR with this test, also with a test for the `create_bonds_with_id` function :+1: 
See #187

roigcarlo: Oh, I see, sorry I was not aware of that PR. We can remove this as soon as #187 is merged

SGGgarcia: @roigcarlo #187 is merged, you can proceed

simphony/testing/abc_check_engine.py
```diff
@@ -35,46 +45,48 @@ def create_dataset_items(self):
...
+    def add_operation(self, container, *args, **kwrds):
+        method = getattr(
+            sys.modules[__name__],
+            self.operation_mapping['add item'])
+        return method(*args, **kwrds)
 
     def test_get_missing_dataset(self):
-        filename = os.path.join(self.temp_dir, 'test.cuds')
-        with closing(H5CUDS.open(filename)) as handle:
+        with closing(self.container_factory('test.cuds')) as handle:
```
nathanfranklin: > This template should be independent of the H5CUDS class.

As mentioned in #182, this needs to be independent of H5CUDS.

I think you need a

simphony/testing/abc_check_engine.py
```diff
@@ -25,6 +30,11 @@ def setUp(self):
         self.items = self.create_dataset_items()
 
     @abc.abstractmethod
+    def container_factory(self, name, mode='w'):
```
nathanfranklin: This is the method to get the engine to be tested, right?  Wouldn't this be called something like "container_engine" and there would be no need for parameters such as `name` or `mode`.

I understand that as there is a overlap with the H5CUDS in that it stores the DataSets like an ABCModelingEngine that these tests are used there too.

simphony/testing/abc_check_engine.py
roigcarlo: Missed that thx

simphony/testing/abc_check_engine.py
roigcarlo: Right I want to obtain the engine here. Maybe even a best option is engine_factory, for containers I understand ```particles```, ```mesh``` and ```lattice```

The problem with `name` and `mode` is that they are actually used in some tests. For example see test: 

https://github.com/simphony/simphony-common/blob/fix-issue-159/simphony/testing/abc_check_engine.py#L234  

There both variations in the name and opening mode are used. Name dosn't need to be there as we can change it inside the test, but if the opening mode is strictly necessary i can't see how to remove that parameter.

Finally, yes, the idea is to use the ```abc_check_engine``` to test all the functionalities related with the datasets of the H5CUDS class

nathanfranklin: The `abc_check_engine.py` is to be used for testing by lammps-md and kratos which follow the ABCModelingEngine class but both don't have a concept of file mode.

I think the confusion is that some of the tests are originally from testing H5CUDS and are testing specific things related to files.  For example, in the current `test_add_get_dataset()` which was from here (lines from version 0.1.5:   https://github.com/simphony/simphony-common/blob/0.1.5/simphony/io/tests/test_h5_cuds.py#L158-161), there is testing related to the behaviour of a H5Particles instance  and the H5Cuds file when the H5CUDS file is closed.  The tests are ensuring that exceptions are thrown when the file has been close (i.e. the user can't use a closed file and the user can no longer use the H5Particles instance which was from that closed file).

Maybe these file-specific tests should only be used in the testing of H5Cuds (i.e. only appear in 'test_h5cuds.py').  The general tests of ABCModelingEngine should be in 'testing/abc_check_engine.py'.

itziakos: I agree, the H5CUDS is a special case.

roigcarlo: Ok, it makes sense now. I have moved the tests back to ```test_h5_cuds.py``` and implemented @nathanfranklin  suggestions.

simphony/testing/abc_check_engine.py
```diff
@@ -35,46 +40,48 @@ def create_dataset_items(self):
...
-        with closing(H5CUDS.open(filename)) as handle:
+        with closing(self.engine_factory()) as handle:
             with self.assertRaises(ValueError):
                 handle.get_dataset('foo')
 
     def test_add_dataset_empty(self):
-        filename = os.path.join(self.temp_dir, 'test.cuds')
-        with closing(H5CUDS.open(filename)) as handle:
-            handle.add_dataset(self.create_dataset(name='test'))
+        with closing(self.engine_factory()) as handle:
```
nathanfranklin: ABCModelingEngine does not have a `close` method;  the use of `closing` from contextlib should be removed.  This is also a hold-over from these tests being from testing H5Cuds.

nathanfranklin: I would suspect that:
```
with closing(self.engine_factory()) as handle:
    handle.add_dataset(reference)
    ...
```
becomes something like
```
handle = self.engine_factory()
handle.add_dataset(reference)
...
```

Then in the specific testing of H5Cuds that uses this (e.g. TestParticlesCudsOperations) would have to be updated to ensure that the proper cleanup occurs (i.e. file is closed)

roigcarlo: Would something like:

```python
handle = self.engine_factory()
...
self.engine_close(handle)
```

be correct? I supose that we have to provide this if we want to tests the FILE-IO wrappers. That's why i left the closing there in the first place. What do you think?

nathanfranklin: > be correct? I supose that we have to provide this if we want to tests the FILE-IO wrappers. That's why i left the closing there in the first place. What do you think?

Any specific cleanup can just be handled in the `tearDown` method of the classes subclassing `CheckEngine`.  So an additonal function 'engine_close' isn't required.

All our wrappers (regardless of being FILE-IO/INTERNAL) implement the ABCModelingEngine interface and with this framework can then be tested in the same manner.   For example, in simphony-lammps-md, we have both a FILE-IO and INTERNAL wrapper and both are tested in exactly the same manner: https://github.com/simphony/simphony-lammps-md/blob/master/simlammps/tests/test_lammps_wrapper.py and https://github.com/simphony/simphony-lammps-md/blob/master/simlammps/testing/abc_lammps_md_engine_check.py.   Once 0.2.0 of simphony-common is released, those tests in simphony-lammps will be changed/extended to use the testing frame work you are developing (i.e. CheckEngine, etc)

simphony/testing/abc_check_engine.py
```diff
@@ -1,21 +1,20 @@
 import abc
 import uuid
-import os
-from contextlib import closing
+import sys
 import tempfile
```
nathanfranklin: tempfile can be removed (and its use in `setUp`)

simphony/io/tests/test_h5_cuds.py
```diff
@@ -17,6 +25,12 @@ class TestH5CUDS(unittest.TestCase):
     def setUp(self):
         self.temp_dir = tempfile.mkdtemp()
         self.maxDiff = None
```
nathanfranklin: maxDiff can be removed

simphony/testing/abc_check_engine.py
```diff
@@ -35,173 +39,186 @@ def create_dataset_items(self):
...
+        'compare datasets': 'none',
+        'add item': 'none'}
+
+    def compare_operation(self, *args, **kwrds):
+        method = getattr(
+            sys.modules[__name__],
+            self.operation_mapping['compare datasets'])
+        return method(*args, **kwrds)
+
+    def add_operation(self, container, *args, **kwrds):
```
nathanfranklin: Correct me if i am wrong, but I think this is unused and this method as well as the "add_items" entry in `operation_mapping` can be removed.

roigcarlo: You are right, this actually was only used by the add_get tests that have been moved to test_h5_cuds.

simphony/testing/abc_check_engine.py
```diff
@@ -35,173 +39,186 @@ def create_dataset_items(self):
...
+        self.compare_operation(reference, ds, testcase=self)
 
     def test_add_dataset_empty_data(self):
-        filename = os.path.join(self.temp_dir, 'test.cuds')
-        with closing(H5CUDS.open(filename)) as handle:
-            handle.add_dataset(self.create_dataset(name='test'))
-            pc = handle.get_dataset("test")
-            self.assertEqual(DataContainer(), pc.data)
-            self.assertEqual(0, len(pc.data))
+        handle = self.engine_factory()
```
nathanfranklin: In looking at `test_add_dataset_empty` above and this 'test_add_dataset_empty_data`.  I don't think these names are accurate on what they are testing.  In general, you just know that `self.create_dataset(name='test')` is providing you with a dataset named "test".  One doesn't know if its empty or if its 'data' is empty.  

Due to that, I would recommend deleting `test_add_dataset_empty_data` completly.   and then rename `test_add_dataset_empty` to `test_add_dataset`

nathanfranklin: >Due to that, I would recommend deleting `test_add_dataset_empty_data` completly.   and then rename `test_add_dataset_empty` to `test_add_dataset`

Oops!  I just see there there is already a `test_add_dataset`.  

I would then suggest just keeping `test_add_dataset` and deleting both  `test_add_dataset_empty` and  'test_add_dataset_empty_data`.

simphony/testing/abc_check_engine.py
```diff
@@ -35,173 +39,186 @@ def create_dataset_items(self):
...
+        handle.add_dataset(self.create_dataset(name='test'))
+        ds = handle.get_dataset("test")
+        data = ds.data
+        data[CUBA.NAME] = 'somename'
+
+        # Since the returned data is always a copy,
+        #  therefore the ds.data should not have changed
+        self.assertNotIn(CUBA.NAME, ds.data)
+        # And the length should be still zero
+        self.assertEqual(0, len(ds.data))
```
nathanfranklin: This test has to be reworked.  It is based on the assumption that the dataset's DataContainer `data` is empty.  

In general, it looks like goal of this test was to check that the getting and setting of `data` of the DataSets  takes place through a copy (#84)

simphony/testing/utils.py
```diff
@@ -9,7 +9,43 @@
 from simphony.core.data_container import DataContainer
 from simphony.core.cuba import CUBA
 from simphony.cuds.particles import Particle, Bond
-from simphony.cuds.mesh import Point
+from simphony.cuds.mesh import Point, Edge, Face, Cell
+
+
+def compare_particles_datasets(particles, reference, msg=None, testcase=None):
```
nathanfranklin: The `msg` parameter is unused in this method (and in subsequent methods).  I would recommend deleting it.

simphony/testing/abc_check_engine.py
```diff
@@ -35,173 +39,186 @@ def create_dataset_items(self):
...
-                    self.create_dataset(name='test'))
+        handle = self.engine_factory()
+        handle.add_dataset(self.create_dataset(name='test'))
+        with self.assertRaises(ValueError):
+            handle.add_dataset(
+                self.create_dataset(name='test'))
 
     def test_iter_dataset(self):
+        handle = self.engine_factory()
         # add a few empty particle containers
```
nathanfranklin: this comment needs to be updated or removed

simphony/testing/abc_check_engine.py
```diff
@@ -35,173 +39,186 @@ def create_dataset_items(self):
...
-            for ds in handle.iter_datasets():
-                handle.remove_dataset(ds.name)
-                # test that we can't get deleted containers
-                with self.assertRaises(ValueError):
-                    handle.get_dataset(ds.name)
-                # test that we can't use the deleted container
-                with self.assertRaises(Exception):
-                    self.check_is_usable_dataset(ds, self.items)
+        handle = self.engine_factory()
+        # add a few empty particle containers
```
nathanfranklin: this comment needs to be updated/deleted

simphony/testing/abc_check_engine.py
```diff
@@ -35,173 +39,186 @@ def create_dataset_items(self):
...
+        for i in xrange(5):
+            name = "test_" + str(i)
+            handle.add_dataset(Particles(name=name))
+
+        # delete each of the particle containers
```
nathanfranklin: this comment needs to be updated/deleted

simphony/testing/abc_check_engine.py
```diff
@@ -35,173 +39,186 @@ def create_dataset_items(self):
...
-            self.assertEqual("bar", ds_bar.name)
+        # we should be able to access using the new "bar" name
+        ds_bar = handle.get_dataset("bar")
+        self.assertEqual("bar", ds_bar.name)
 
-            # and we should be able to use the no-longer used
-            # "foo" name when adding another particle container
-            ds = handle.add_dataset(self.create_dataset(name='foo'))
+        # and we should be able to use the no-longer used
+        # "foo" name when adding another particle container
```
nathanfranklin: this comment needs to be updated/deleted

simphony/io/tests/test_h5_cuds.py
```diff
@@ -78,15 +178,72 @@ def test_incorrect_version(self):
...
 
 class TestParticlesCudsOperations(ParticlesCudsCheck, unittest.TestCase):
-    pass
+
+    def setUp(self):
+        ParticlesCudsCheck.setUp(self)
+        self.engines = []
+
+    def engine_factory(self):
+        filename = os.path.join(self.temp_dir, 'test.cuds')
```
nathanfranklin: Once, https://github.com/simphony/simphony-common/pull/189/files#r34765387 is addressed, this needs to be looked at to ensure that proper cleanup of the temp file/directories occurs

nathanfranklin: Same applies to `TestMeshCudsOperations` and `TestLatticeCudsOperations`

simphony/testing/abc_check_engine.py
nathanfranklin: I think this should be changed somehow.  

I understand that `create_dataset` returns either an ABCMesh, ABCParticles or ABCLattice and then you need to know which compare function to use (e.g. `compare_particles_datasets`...). 

 Is there another way that ParticlesCudsCheck, LatticeCudsCheck and MeshCudsCheck provide this information (e.g. an abstractmethod which they each implement to provide this information)?

simphony/testing/abc_check_engine.py
```diff
@@ -292,68 +256,15 @@ def create_dataset_items(self):
...
 
-    def check_emtpy_dataset(self, ds):
-        """ Check if a dataset is empty
-        """
 
-        self.assertEqual(0, len(list(p for p in ds.iter_points())))
-        self.assertEqual(0, len(list(e for e in ds.iter_edges())))
-        self.assertEqual(0, len(list(f for f in ds.iter_faces())))
-        self.assertEqual(0, len(list(c for c in ds.iter_cells())))
+class LatticeCudsCheck(CheckEngine):
```
nathanfranklin: This is a suggestion:  

Could `LatticeCudsCheck`, `ParticlesCudsCheck` and `MeshCudsCheck` be renamed to reflect that they are checking how an engine supports that type of DataSet.    What about something like "LatticeEngineCheck" or "EngineLatticeCheck"?

nathanfranklin:
@roigcarlo ,  the testing for ABCModelingEngine is looking good.  Just left some comments to improve some things.  

The only thing other thing is that we currently don't have anything that tests if the engine's BC, CM or SP DataContainers exist.  It might be that others disagree but my initial feeling is that maybe that doesn't matter at this point (and that it is fine that such tests don't currently exist)

roigcarlo:
@nathanfranklin Thx for the feedback, I believe I adressed all the issues plus some others I found.

Regarding the BC, CM and SP DataContainers, it is true that we do not have any test. To be honest I did not implemented them becase I do not have any current code to run the tests against, I suppose that once either lammps or kratos use this interface we can add it in a separate PR

simphony/testing/abc_check_engine.py
```diff
@@ -35,173 +35,191 @@ def create_dataset_items(self):
...
+        data = ds.data
+        data[CUBA.NAME] = 'somename'
+        data[CUBA.MATERIAL_ID] = 0
+
+        # Since the returned data is always a copy,
+        #  therefore the ds.data should not have changed
+        self.assertNotEqual(data[CUBA.NAME], ds.data[CUBA.NAME])
+        self.assertNotIn(CUBA.MATERIAL_ID, ds.data)
+        # And the length should be still one
+        self.assertEqual(1, len(ds.data))
```
nathanfranklin: I think this assert and the other assert related to this size of ds.data (about 10 lines down) should both be removed.  As we don't know what is in ds.data (as it originally came from somewhere else: ```self.create_dataset(name='test')```)

nathanfranklin:
@roigcarlo , just left two minor changes otherwise looks great. 

I think the only open point is: https://github.com/simphony/simphony-common/pull/189#discussion-diff-34590829

also, my suggestion to delete maxDiff was wrong; I just realized that its an attribute of unittest to control maximum length of diffs. Sorry about that.

simphony/testing/abc_check_engine.py
roigcarlo: I belive we can keep those asserts if we just query the size of the original DataContainer, something like:

```python 
reference = self.create_dataset(name='test')
reference_data = reference.data
reference_datasize = len(reference_data)

...

self.assertEqual(reference_datasize , len(ds.data))

...

self.assertEqual(reference_datasize + 1, len(ds.data))

```

Along with the other tests it should guarantee that the data is the same and no extra data has been added in the process. This is important since, as you mentioned we cannot relay on the data inside the datatest being empty ( it may happens that it already contains CUBA.MATERIAL_ID).

Furthermore, I would also suggest to use our own DataContainer created inside the test, but this conflicts somehow with the purpose of the test.

nathanfranklin: Yeah, that could work. 

I only  wonder if just the checks with one CUBA_keyworkd (like 'CUBA.NAME') are sufficient to check that getting/setting of `data` is taking place through a copy.

```        
       handle = self.engine_factory()

        reference = self.create_dataset(name='test')

        reference_data = reference.data
        reference_data[CUBA.NAME] = 'foo_name'
        reference.data = reference_data

        handle.add_dataset(reference)

        ds = handle.get_dataset('test')

        data = ds.data
        data[CUBA.NAME] = 'somename'

        # Since the returned data is always a copy,
        #  therefore the ds.data should not have changed
        self.assertNotEqual(data[CUBA.NAME], ds.data[CUBA.NAME])

        #if
        ds.data = data
        
        #then
        self.assertEqual(data[CUBA.NAME], ds.data[CUBA.NAME])
```

That is just an idea though.  Anyway is fine with me.

roigcarlo: @ I have left what you suggested above. We are always in time to extend in a future PR if some more tests are needed.

PR #205, opened 2015-08-18
PR: Feature lattice prim cell

kemattil:
Define Bravais lattices using a primitive cell.

Main features (files ```primitive_cell.py```, ```abc_lattice_new.py```, and ```lattice_new.py```):
- a new data structure for representing primitive cells of a lattice (class ```PrimitiveCell```)
- ```ABCLattice``` api changed, i.e. properties ```type``` and ```base_vect``` replaced with ```prim_cell```
- standalone class ```Lattice``` takes a reference to ```PrimitiveCell``` at the initialization
- ```get_coordinates``` implemented in ```ABCLattice```; uses the primitive vectors from ```PrimitiveCell``` (properties ```p1```, ```p2```, and ```p3```)
- the factory functions ```make_XXX_lattice``` call the ```PrimitiveCell``` class methods ```create_cell_XXX_lattice```
- the ```H5Lattice``` api remains almost intact, i.e. class method ```create_new``` should take ```prim_cell``` instead of ```type``` and ```base_vect```, and the primitive cell should be read and stored in I/O


tuopuu:
@kemattil, looks good. It's a surprisingly simple implementation, which is a good thing. (... but it doesn't have quite as many features as the competing one, however.) I'll give my vote to this.

One comment for the code already written: we should not use C-style format for floating point numbers: just 0 or 2, instead of 0.0 and 2.0.

If other guys from the dev-team prefer this proposal, I can take over and start writing tests, h5lattice, etc.

simphony/cuds/primitive_cell.py
```diff
@@ -0,0 +1,537 @@
...
+        self._p3 = p3
+
+        if np.isclose(self.volume(), 0):
+            message = 'Cell volume must be non-zero'
+            raise ValueError(message)
+
+        self._bravais_lattice = BravaisLattice.TRICLINIC
+            
+    @classmethod
+    def create_cell_cubic_lattice(cls, a):
```
itziakos: There is no need to repeat the word `cell` in the methods

simphony/cuds/primitive_cell.py
```diff
@@ -0,0 +1,537 @@
...
+    ----------
+    p1, p2, p3: 3 x 3D vector
+        primitive vectors
+
+    bravais_lattice: BravaisLattice(Enum)
+        the 3D Bravais lattice for which the primitive cell is defined,
+        i.e. not the symmetry group or type of the primitive cell itself
+    """
+   
+    def __init__(self, p1, p2, p3):
```
itziakos: You can have an optional argument for the lattice kind that is set to `None`. This will remove the uncomfortable access of the `self._bravais_lattice` in the class methods and the arbitrary value in line 82.

simphony/cuds/primitive_cell.py
itziakos: Looking at the methods I would suggest to rename the `classmethods` just with the name of the cell e.g:
```
def cubic(cls, a)
def body_centered_cubic(cls)   # we should avoid cutting the words short
```
This will make the functions short and easier to understand

```
cell = PrimitiveCell.cubic(10)
```

tuopuu: I would get rid of the lattice word instead of cell.

simphony/cuds/lattice_new.py
```diff
@@ -0,0 +1,342 @@
...
+    org : float[3], default value = (0, 0, 0)
+        lattice origin
+
+    Returns
+    -------
+    lattice : Lattice
+        A reference to a Lattice object.
+    """
+    pcell = pc.create_cell_cubic_lattice(h)
+    return Lattice(name, 'Cubic', size, org, pcell)
```
itziakos: I think that these need to be updated with the new constructor of the Lattice class, there is no need to pass the lattice kind anymore

simphony/cuds/primitive_cell.py
```diff
@@ -0,0 +1,537 @@
+import numpy as np
+from enum import Enum
+
+
+class BravaisLattice(Enum):
```
itziakos: better subclass `IntEnum`

simphony/cuds/primitive_cell.py
itziakos: I think we can remove `cell` also. The info is in the class name

itziakos: perfect

itziakos:
So far it looks very good and concise. Had only minor comments. Looking forward for the complete PR

simphony/cuds/primitive_cell.py
kemattil: We have to be careful with the class method names: the methods create primitive cells for specific bravais lattices.

For example the specified primitive cell for a body-centered cubic lattice can itself be classified as a rhombohedral cell (i.e. it belongs to the rhombohedral symmetry group).

Therefore, a line ```cell = PrimitiveCell.body_centered_cubic(10)``` might be misleading ...

A line ```cell = PrimitiveCell.body_centered_cubic_lattice(10)``` appears more precise.
Note that the argument ```10``` specifies a (conventional) unit cell edge length, i.e.
the argument specifies a body-centered cubic lattice for which a primitive cell is created.

Would this be an acceptable compromise?

itziakos: The main problem is that these functions do not return a `lattice`.

Is there a unique name for this kind of `lattice`.
Do the other `body_centered_cubic` types need only one argument?

For the sake of user experience and a small cost in complexity we can support multiple signatures for `body_centered_cubic` so that more that one type of PrimaryCells can be returned.

kemattil: I'm not sure that I understand the questions, but the ```body_centered_cubic``` describes
a lattice type, i.e. there are no other ```body_centered_cubic``` types.

But there are infinitely many ```body_centered_cubic``` lattices and, moreover,
for any given lattice of this type, the primitive cell is not unique.

For example, the current implementation ```PrimitiveCell.body_centered_cubic```
returns a specific primitive cell for a specific body-centered cubic lattice.

To allow construction of arbitrary primitive cells for arbitrary lattices of
a given type is a huge undertaking (I don't know how to do it).

On the other hand, the current initialization (i.e. ```__init__```)provides the most general way of specifying primitive cells. However, at the initialization, the lattice type is designated
as ```triclinic``` by default (all other lattice types are special cases of ```triclinic``` lattices).

Currently the initialization does not allow users to specify the lattice type, because
we do not know how to check whether the provided primitive vectors really
define a lattice of the given type (a very non-trivial task) ...

kemattil: Is ```cell = PrimitiveCell.for_body_centered_cubic_lattice(10)``` better?

simphony/cuds/primitive_cell.py
kemattil: This is related to my previous comment. We do not know which approach is the best compromise between safety and versatility/simplicity ...

simphony/cuds/lattice_new.py
kemattil: Corrected.

simphony/cuds/primitive_cell.py
kemattil: Changed.

kemattil:
@tuopuu , I have made the small changes proposed last week (consensus on the ```PrimitiveCell``` class method names or signature of ```PrimitiveCell.__init__``` has not yet been achieved).

Can you continue from here? Note that the code is absolutely not tested ...

tuopuu:
> Can you continue from here? Note that the code is absolutely not tested ...

@kemattil I'll continue from here.

tuopuu:
I have updated the tests for primitive vector representation of the Lattice and H5Lattice classes. Old H5Lattice implementation was only modified slightly, so that we still use Pytables attributes of the table structure to store size,origin and primitive vectors of the lattice. Because of the file backend changes, I have changed the LATTICE_CUDS_VERSION to 2. 

Please note, that there is no conversion tool for H5CUDS files available for this change: we need to decide where to put these tools in the SimPhoNy file structure (maybe simphony/utils?). (Here, writing the conversion script is trivial, however.)

Please also note, that documentation is not up to date, and that Appveyor build fails. I don't know why. It should have nothing to do with this branch...

@kemattil, @roigcarlo, @nathanfranklin, this branch is ready for review. Please take a look.

simphony/io/h5_lattice.py
```diff
@@ -28,20 +29,22 @@ def __init__(self, group):
             raise ValueError("Lattice file layout has an incompatible version")
 
         self._group = group
-        self._type = group.lattice.attrs.type
-        self._base_vect = group.lattice.attrs.base_vect
+        self._prim_cell = PrimitiveCell(group.lattice.attrs.prim_cell[0],
+                                        group.lattice.attrs.prim_cell[1],
+                                        group.lattice.attrs.prim_cell[2])
+        self._prim_cell._bravais_lattice = BravaisLattice(
```
nathanfranklin: In order to avoid using the non-public `_bravais_lattice`  attribute here, I would suggest possibly reworking PrimitiveCell so that `bravais_lattice` is also a parameter in PrimitiveCell's ` __init__`

simphony/cuds/abc_lattice.py
```diff
@@ -8,17 +8,14 @@ class ABCLattice(object):
     ----------
     name : str
         name of lattice
-    type : str
-        type of lattice
-    base_vect : float[3]
-        base vector of lattice
+    prim_cell : PrimitiveCell
```
nathanfranklin: This is just a suggestion that can be ignored, but I prefer the attribute name `primitive_cell` instead of `prim_cell`

dpinte: +1

simphony/cuds/lattice.py
```diff
@@ -190,66 +155,180 @@ def data(self, value):
...
     Parameters
     ----------
     name : str
     h : float
         lattice spacing
-    size : int[2]
+    size : int[3]
         Number of lattice nodes in each axis direction.
-    origin : float[3], default value = (0, 0, 0)
+    org : float[3], default value = (0, 0, 0)
```
nathanfranklin: suggestion: `org` -> `origin`

tuopuu: You're right, it's better not to use abbreviations. It was changed to avoid line shifts but that's obviously not a big deal.

simphony/io/h5_lattice.py
tuopuu: Good idea. PrimitiveCell can be modified to include bravais_lattice parameter. It will also simplify some of the classmethods.

kemattil:
@tuopuu, I think this PR if fine.

nathanfranklin:
:+1:

tuopuu:
Added tests for PrimitiveCell class method parameters so that proper error are returned when the values are not on the allowed range. Also removed the test for cell volume greater than zero in the PrimitiveCell constructor, because that test always passes when edge length and edge parallelity tests pass.

I think this PR is ready to be merged, but waiting for one +1 for the final additions.

kemattil:
+1

PR #207, opened 2015-08-24
Setup appveyor runs

itziakos:
This PR adds setup files to run tests on appveyor (windows). Currently the hdf5 based tests are skipped because building the hdf5 package for python 2.7 is a little tricky.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,8 +17,8 @@ init:
   - ps: $Env:sdkverpath = "C:/Program Files/Microsoft SDKs/Windows/" + $Env:sdkver + "/Setup/WindowsSdkVer.exe"
   - ps: $Env:path = $Env:python + ";" + $Env:python + "\scripts;" + $Env:sdkbin + ";" + $Env:path
 install:
-  - ps: (new-object net.webclient).DownloadFile('https://raw.github.com/pypa/pip/master/contrib/get-pip.py', 'c:/get-pip.py')
-  - ps: python c:/get-pip.py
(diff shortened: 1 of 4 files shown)


itziakos:
added issue #208 for the hdf5 based tests

appveyor.yml
```diff
@@ -0,0 +1,25 @@
...
+
+cache:
+  - '%localappdata%\pip\cache'
+
+init:
+  - ps: $Env:sdkbin = "C:\Program Files\Microsoft SDKs\Windows\" + $Env:sdkver + "\Bin"
+  - ps: $Env:sdkverpath = "C:/Program Files/Microsoft SDKs/Windows/" + $Env:sdkver + "/Setup/WindowsSdkVer.exe"
+  - ps: $Env:path = $Env:python + ";" + $Env:python + "\scripts;" + $Env:sdkbin + ";" + $Env:path
+install:
+  - ps: (new-object net.webclient).DownloadFile('https://raw.github.com/pypa/pip/master/contrib/get-pip.py', 'c:/get-pip.py')
```
tuopuu: Should the copy path for the file get-pip.py be something people generally have better user rights, e.g. C:/Temp? This would be a problem in my work computer where I only have very limited rights.

tuopuu:
@itziakos, one comment, but I don't think I can properly review this one, so I suggest to wait someone else to also give their feedback.

appveyor.yml
itziakos: I will make the change.

Please note this setup is specific for appveyor continuous testing (http://www.appveyor.com) which uses Windows VM and the discards them, so in that case it does not really matter where we save the `get-pip.py` bootstrap code.

roigcarlo:
As far as I can tell :+1:

SGGgarcia:
Looks ok to merge :+1:

Issue #219, opened 2015-11-05
problem using test_delete_dataset and test_iter_dataset in CheckEngine


nathanfranklin:
`test_delete_dataset` uses ds.name after ds has been deleted.  It also remove datasets while iterating over the datasets.

`test_iter_dataset` assumes that  `iter_dataset` returns the datasets in the order that they were added.

nathanfranklin:
Closed by #220

tuopuu:
I'll reopen this issue, because ```test_delete_dataset``` still references the proxy object and therefore the object is not deleted. See my previous line comment.

nathanfranklin:
@tuopuu , we can close this again, right?

tuopuu:
Yes, we can close this.

PR #220, opened 2015-11-09
Fix engine tests

nathanfranklin:
This PR addresses #219

### DIFF
--- a/simphony/testing/abc_check_engine.py
+++ b/simphony/testing/abc_check_engine.py
@@ -152,8 +152,7 @@ class CheckEngine(object):
         # test iterating over all
         names = [
             ds.name for ds in engine.iter_datasets()]
-        self.assertEqual(len(names), len(ds_names))
-        self.assertEqual(set(names), set(ds_names))
+        self.assertItemsEqual(names, ds_names)
 
         # test iterating over a specific subset
         subset = ds_names[:3]
--- a/simphony/testing/abc_check_engine.py
+++ b/simphony/testing/abc_check_engine.py
@@ -152,7 +152,8 @@ class CheckEngine(object):
         # test iterating over all
         names = [
             ds.name for ds in engine.iter_datasets()]
-        self.assertEqual(names, ds_names)
+        self.assertEqual(len(names), len(ds_names))
+        self.assertEqual(set(names), set(ds_names))
 
         # test iterating over a specific subset
         subset = ds_names[:3]
--- a/simphony/testing/abc_check_engine.py
+++ b/simphony/testing/abc_check_engine.py
@@ -178,15 +178,17 @@ class CheckEngine(object):
             name = "test_" + str(i)
             engine.add_dataset(self.create_dataset(name=name))
 
+        datasets_and_names = [(ds, ds.name) for ds in engine.iter_datasets()]
+
         # delete each of the datasets
-        for ds in engine.iter_datasets():
-            engine.remove_dataset(ds.name)
+        for (dataset, name) in datasets_and_names:
+            engine.remove_dataset(name)
             # test that we can't get deleted datasets
             with self.assertRaises(ValueError):
-                engine.get_dataset(ds.name)
+                engine.get_dataset(name)
             # test that we can't use the deleted datasets
             with self.assertRaises(Exception):
-                self.compare_dataset(ds, ds)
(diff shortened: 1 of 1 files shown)


tuopuu:
Could you think of a way of testing whether or not the engine being tested supports adding multiple datasets in test_iter_datasets and test_remove_dataset, and according to the results of that test, either 5 or 1 datasets are initiated in these two tests?

Otherwise this PR is :+1:

nathanfranklin:
> Could you think of a way of testing whether or not the engine being tested supports adding multiple datasets in test_iter_datasets and test_remove_dataset, and according to the results of that test, either 5 or 1 datasets are initiated in these two tests?

I'll add a separate issue for this.

nathanfranklin:
@tuopuu , @roigcarlo , good to merge?

roigcarlo:
Looks good to me :+1:

simphony/testing/abc_check_engine.py
```diff
@@ -177,15 +177,17 @@ def test_delete_dataset(self):
...
+        for (dataset, name) in datasets_and_names:
+            engine.remove_dataset(name)
             # test that we can't get deleted datasets
             with self.assertRaises(ValueError):
-                engine.get_dataset(ds.name)
+                engine.get_dataset(name)
             # test that we can't use the deleted datasets
             with self.assertRaises(Exception):
-                self.compare_dataset(ds, ds)
+                self.compare_dataset(dataset, dataset)
```
tuopuu: This line still causes an error, because ```dataset``` is referencing to the proxy object, and therefore the proxy object is not deleted.

I don't see a clean way how this could be fixed, other than removing this part of the test completely. In Python the object is deleted when there is no references to it, and even trying this test here requires that there is a reference existing.

Maybe compare_dataset could be modified to check if the proxy object has a ```_am_i_deleted``` attribute set to True, and in that case return an exception, but this doesn't sound a very elegant solution.

nathanfranklin: I think the test is okay as it is.

Lines 189-190 is to check that we get an error when we try to use the deleted `dataset`.   From line 180, we know that  `dataset` is a *proxy of the dataset named name that is stored internally in the Engine.*  In line 184, we know that the dataset with this name was removed from the engine.  Therefore, the proxy 'dataset' should no longer work.  It should throw an exception if it is used (i.e. data.name, data.iter_nodes ... or other things that will be called in `self.compare_dataset`).

tuopuu: I tested it with Jyulb wrapper and this particular line causes a Segmentation Fault without any other messages. When I check the type of ```dataset``` before the compare_datasets call, it reports a ProxyLattice object.

nathanfranklin: Yeah, JYULB wrapper fails at the moment for this test.   I would suggest opening up an issue in simphony-jyulb to follow it and override the method in your subclass of LatticeEngineCheck.

We want something like the following to happen:

```
foo_proxy_lattice = jyulb_engine.get_dataset("foo")
for node in foo_proxy_lattice.iter_nodes():
     print node.data[CUBA.VELOCITY]
jyulb_engine.remove_dataset("foo")
print(type(foo_proxy_lattice))  # prints <class 'jyulb....ProxyLattice''>
for node in foo_proxy_lattice.iter_nodes():  # this should raise an exception
```

tuopuu: I don't think this error is related to JYULB wrapper specifically. Have you tried this test for any other wrapper? I think it will always fail. If I'm mistaken, then sure we can make a new issue for it and try to solve it there.

nathanfranklin: It works for `simphony-lammps-md` (https://github.com/simphony/simphony-lammps-md/blob/master/simlammps/tests/test_lammps_wrapper.py).

I take a look at simphony-jyulb in the next few days.

-------------

>  I tested it with Jyulb wrapper and this particular line causes a Segmentation Fault without any other messages. When I check the type of dataset before the compare_datasets call, it reports a ProxyLattice object. 

Can you recreate this (i.e. have some code that causes the segmentation fault)?   Why don't you debug it too see when the segmentation fault occurs. 

Lets discuss this on a different issue in simphony-jyulb.

Issue #221, opened 2015-11-09
Extend TestEngine so that it can be used for "single-dataset" engines


nathanfranklin:
In `TestEngine`,  `test_iter_datasets` and `test_remove_dataset` add 5 datasets for testing purposes.  Some engines only support a **single** dataset.

nathanfranklin:
`TestEngine` could be given given a method that could be overridden for testing *single-dataset* engines

something like:
```
    def number_datasets_used_in_testing(self):
        """ Number of datasets used in testing engine

        """
        return 5
```

nathanfranklin:
Closed by #224

PR #224, opened 2015-11-10
Extend tests to allow certain number of datasets

nathanfranklin:
This PR addresses #221.  Some engines support only certain number of datasets (e.g. single-dataset engines).  This PR allows the number of datasets used during testing to be set.   For example, tests for an engine that only supported a single-lattice would look like the following.:

```
class MySingleLatticeEngineCheck(LatticeEngineCheck):
    def setUp(self):
        LatticeEngineCheck.setUp(self, number_datasets_used_in_testing=1)

```

### DIFF
--- a/simphony/testing/abc_check_engine.py
+++ b/simphony/testing/abc_check_engine.py
@@ -215,6 +215,7 @@ class CheckEngine(object):
         # we should be able to access using the new "bar" name
         ds_bar = engine.get_dataset("bar")
         self.assertEqual(ds_bar.name, "bar")
+        engine.remove_dataset("bar")
 
         # and we should be able to use the no-longer used
         # "foo" name when adding another dataset
--- a/simphony/testing/abc_check_engine.py
+++ b/simphony/testing/abc_check_engine.py
@@ -215,7 +215,8 @@ class CheckEngine(object):
         # we should be able to access using the new "bar" name
         ds_bar = engine.get_dataset("bar")
         self.assertEqual(ds_bar.name, "bar")
-        engine.remove_dataset("bar")
+        if self.number_datasets_used_in_testing < 2:
+            engine.remove_dataset("bar")
 
         # and we should be able to use the no-longer used
         # "foo" name when adding another dataset
--- a/simphony/testing/abc_check_engine.py
+++ b/simphony/testing/abc_check_engine.py
@@ -15,9 +15,10 @@ class CheckEngine(object):
 
     __metaclass__ = abc.ABCMeta
 
-    def setUp(self):
+    def setUp(self, number_datasets_used_in_testing):
         self.maxDiff = None
         self.items = self.create_dataset_items()
+        self.number_datasets_used_in_testing = number_datasets_used_in_testing
 
     @abc.abstractmethod
     def engine_factory(self):
@@ -129,7 +130,7 @@ class CheckEngine(object):
         # add a few empty datasets
         ds_names = []
 
-        for i in xrange(5):
+        for i in xrange(self.number_datasets_used_in_testing):
             name = "test_{}".format(i)
             ds_names.append(name)
             engine.add_dataset(self.create_dataset(name=name))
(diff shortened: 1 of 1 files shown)


nathanfranklin:
This is ready for review.

@tuopuu ,  will this address #221?

tuopuu:
@nathanfranklin, I'm testing this with the new version of JYULB wrapper, and trying to get rid of the overridden tests. So far I haven't been able to make it work though. I'll try to find the reason for it tomorrow.

tuopuu:
Ok, this modification will do. :+1: However, there's still an issue with #219 which should be dealt before I can remove the overridden tests.

tuopuu:
There is a problem with CheckEngine.test_dataset_rename method. We should put 
```
        if self.number_datasets_used_in_testing == 1:
            engine.remove_dataset("bar")
```
in front of the last line in that method.

tuopuu:
@nathanfranklin, I made the change I proposed (without the ```if``` statement). Looks OK to me now.

simphony/testing/abc_check_engine.py
```diff
@@ -212,6 +215,7 @@ def test_dataset_rename(self):
         # we should be able to access using the new "bar" name
         ds_bar = engine.get_dataset("bar")
         self.assertEqual(ds_bar.name, "bar")
+        engine.remove_dataset("bar")
```
nathanfranklin: @tuopuu , I don't think we need this.   We have renamed the "foo" to "bar".   So now in line 222, we can add a dataset with the name of "foo".

tuopuu: The point is that engines that only support a single dataset will not allow adding another dataset with the name "foo" (or with any other name either).

nathanfranklin: Ah, okay.  I see your point now.  I prefer the first solution you proposed using the if-statement. Something like:
```
if self.number_datasets_used_in_testing > 1:
   # and we should be able to use the no-longer used
   # "foo" name when adding another dataset 	# "foo" name when adding another dataset
   ds = engine.add_dataset(self.create_dataset(name='foo')) 	
   ds = engine.add_dataset(self.create_dataset(name='foo'))
```

tuopuu: We can add the if statement there, but this test method it not really testing the number of datasets that are allowed to be stored, only renaming, so we could make it independent of ```number_datasets_used_in_testing```. Adding of more than one datasets is tested elsewhere for engines that support it....

Just wanted to underline this, but anyway, either way is good for me.

tuopuu:
@nathanfranklin I added the if statement. Let's merge this now so we can see how it affects the JYU-LB wrapper tests. The SegFault error might be specific to my system, but I'm quite sure the error will occur in Travis run as well.

simphony/testing/abc_check_engine.py
nathanfranklin: Yeah, the test is about renaming these proxy datasets.  After we rename the dataset from "foo" to "bar", we do some testing that can be broken in 3 parts:

 * 1st part: Then we test that we can't get a "foo" named database (cause its been renamed to "bar").  
 * 2nd part:  we can use the "bar" name
 * 3rd part:  finally, we test if we can add a new dataset called "foo".  We should be allowed to do this as there is only a "bar" dataset at the moment. 

**Only** this final step ("3rd part") is there a problem for wrappers that only allow a single dataset as they can't add a second dataset to test this behaviour.  So that is why I am proposing to add the if-statement there to skip this 3rd part for single-dataset wrappers.

-------------------

If we use 1ad0dfb , there is a problem as we are are deleting the "bar" dataset which weakens the point of the 3rd part of this step.  If we leave this commit (1ad0dfb), the 3rd part ends up just testing if we can add a dataset to wrapper which has no datasets.  This doesnt match up with the original intention of testing if we can add a data set with the 'no-longer used "foo" name`

nathanfranklin:
@tuopuu , i moved the if statement to be able to remove the use `remove_dataset`

tuopuu:
Ok, this looks better. :+1:

Issue #218, opened 2015-11-05
get_dataset_names implemenation/documentation is inconsistent


nathanfranklin:
#200 implemented in #206 added the method `get_dataset_names` 

the method is documented as follows in `ABCModelingEngine`:
```
def get_dataset_names(self):
	""" Returns the names of the all the datasets in the engine workspace.
```

But the testing framework, implementation in H5CUDS and in the UML diagrams treats the method as returns an iterator.

nathanfranklin:
I think we should switch the implementation and documentation so that `get_dataset_names` returns a list of names.

nathanfranklin:
Closed by #230

PR #230, opened 2015-11-17
Fix issue 218

roigcarlo:
This should implement #218.

Please do not take into account changes in the first and the last commit. I had some problems with linter executing flake8 with draconian flags...

Apart from that it should be good to go once checks finalize.


nathanfranklin:
:+1:

PR #234, opened 2015-11-24
Fix docstring when merging #231

roigcarlo:
Pr to fix the problems with last merge (#231)

### DIFF
--- a/simphony/io/h5_cuds.py
+++ b/simphony/io/h5_cuds.py
@@ -36,7 +36,7 @@ class H5CUDS(object):
 
     @classmethod
     def open(cls, filename, mode="a", title='', filters=None):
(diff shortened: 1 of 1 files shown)


nathanfranklin:
:+1: 

---------
@roigcarlo ,  the date in your commits here and in #231 seem to be off.  I think your time zone setting might be off on your dev machine.

roigcarlo:
@nathanfranklin I see the dates correct. Last commit appears to be on nov 24, approx one hour ago in my machine. Perhaps there is some problem with the summer time... I will investigate that thx.

roigcarlo:
@nathanfranklin Well... it looks like the last update set my system time to in GMT + 0 so basically all times I was looking at in GitHub were also in GMT. It should be fixed now. thx!!

Issue #227, opened 2015-11-16
add Material


nathanfranklin:
In relation with #226 (https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/MaterialRelations_proposal), a `Material` class has been defined by the SSB as the following:

```
class Material(object):
    """A class that represents a material.

    Class describes a material and its data.

    Attributes
    ----------
    description : str
        textual description of the material
    uid : uuid.UUID
        the id of the material.
    data : DataContainer
        data associated with this material

    """
```
And the the `uid` of materials will be used like the following:
```
material_foo = Material(name="foo")
material.data[CUBA.MASS] = 1.0  # this is a common mass for entities of this material type

p = Particle()
p.data[CUBA.MATERIAL] = material_foo.uid

engine.SD.add_material(material_foo)

dataset = engine.get_dataset("some_name")
dataset.add_particles([p])
```

This requires: 
- [ ] Adding a new attribute (CUBA.MATERIAL) with type that can represent the UUID
- [ ] Adding `Material` class (plus tests and documentation)
- [ ] Add `SD` object to contain these materials (i.e. `add_material`, `get_material`, `delete_material`, `update_material`) (plus tests and documentation)

ahashibon:
It would be better at this stage to hold on with the SD component, as it may not be needed at this stage. Implementing the Material in the engine now is sufficient. This is due to ongoing discussions in the Standards Body regarding the form of the CUDS.

nathanfranklin:
@ahashibon , sounds good.  I'll update #237 so that the methods to work with materials are added to ABCModeling engine i.e:
```
material_foo = Material(name="foo")
material.data[CUBA.MASS] = 1.0 
engine.SD.add_material(material_foo)
```

nathanfranklin:
>It would be better at this stage to hold on with the SD component, as it may not be needed at this stage. Implementing the Material in the engine now is sufficient. This is due to ongoing discussions in the Standards Body regarding the form of the CUDS.  

@ahashibon   Should we also hold off on the SP component in  #226 as well?   Should we stop extending the SP object to contain these material relations and instead just add the material-relations methods (i.e. add_material_relation, get_material_relation, delete_material_relation, update_material_relation) to `ABCModelingEngine`

ahashibon:
[...]

Issue #228, opened 2015-11-17
Extend H5CUDS so that it can store CUDS Data


roigcarlo:
Extend H5CUDS to be able to save material.  In #227 `Material` class will be created.  These materials will be maintained in an `SD` class (e.g. `engine.SD.add_material`).  H5CUDS should be extended so that either this SD or just the materials themselves can be "saved to"/"read from" a file.

- [ ] Extend H5CUDS to be able to save materials
- [ ] Documentation
- [ ] Tests

roigcarlo:
I will move this to 0.3.1 in order to make the release 0.3.0 for integration

roigcarlo:
Closing since MR is now part of metaclass and hance is not longer relevant

PR #237, opened 2015-11-25
Feature material

nathanfranklin:
This PR addresses #227 
* [x] add Material class
* [x] ~~add StateData to hold material class~~
* [x] add 'Materials' to hold different materials
* [x] add 'Materials' to 'ABCModelingEngine' to store materials
* [x] add CUBA.MATERIAL
* [x] add testing
* [x] add documentation
* [x] fix current workaround in 1aac887 for reading/writing CUBA.Material to file


simphony/cuds/abc_modeling_engine.py
```diff
@@ -16,6 +16,8 @@ class ABCModelingEngine(object):  # pragma: no cover
         container of attributes related to the computational method
     SP : DataContainer
         container of attributes related to the system parameters/conditions
+    materials : Materials
```
nathanfranklin: There is some open question on where the materials should be stored (and how they should be accesed).   Right now,  I have the following
```
material_foo = Material(name="foo")
material.data[CUBA.MASS] = 1.0 
engine.materials.add_material(material_foo)
```
See #227 and https://github.com/simphony/simphony-common/issues/227#issuecomment-159869043

Any thoughts/ideas/critique would be appreciated.

nathanfranklin:
This is ready for review. 

Note the earlier comment:https://github.com/simphony/simphony-common/pull/237#discussion_r45990165

tuopuu:
@nathanfranklin, I can review this later today or tomorrow. (This should be merged for #228 work to start)

doc/Makefile
```diff
@@ -181,4 +181,4 @@ uml:
 	java -jar plantuml.jar "source/uml/" "source/images"
 	mv source/uml/*.png source/images
 	@echo
-	@echo "Coversion finished the uml files are in source/images."
+	@echo "Cocversion finished the uml files are in source/images."
```
tuopuu: Should be "conversion"

doc/source/cuds_definition.rst
```diff
@@ -89,3 +89,14 @@ Core items
...
+Materials
+----------
+
+.. figure:: ./images/material_materials.png
+
+   **Figure 4:** UML diagram of Materials and Material.
+
+In the SimPhoNy modeling framework, :class:`~.Materials` and :class:`~.Material`
+are used to describe materials and to define common material properties. Each
+low-level  (e.g. Point, Particle) can be of a certain material.
```
tuopuu: Maybe "Each low-level _object_ ..."

simphony/cuds/material.py
```diff
@@ -0,0 +1,31 @@
...
+class Material(object):
+    """A class that represents a material.
+
+    Class describes a material and its data.
+
+    Attributes
+    ----------
+    description : str
+        textual description of the material
+    uid : uuid.UUID
```
tuopuu: A minor thing, but maybe the attributes list here should be in the same order as in the uml image (material.txt).

simphony/cuds/abc_modeling_engine.py
tuopuu: I think this is a good way to proceed. I don't see a point in using too much time on this, because there will be a change coming to it anyway.

simphony/cuds/material.py
tuopuu: Or better, make the figure and docstrings here follow the order or arguments in the __init__ method.

simphony/cuds/materials.py
```diff
@@ -0,0 +1,99 @@
+from simphony.cuds.material import Material
+
+
+class Materials(object):
+    """A class that contains materials.
+
+    Class manages the materials.
+
+    """
+    def __init__(self, uid=None, data=None, description=""):
```
tuopuu: All arguments here seem to be unused.

simphony/cuds/materials.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        del self._materials[uid]
+
+    def update_material(self, material):
+        """ Update a material
+
+        Parameters
+        ----------
+        material : Material
+            the material to be added.
```
tuopuu: Maybe change "added" to "updated" ...

simphony/cuds/material.py
```diff
@@ -0,0 +1,31 @@
...
+        the id of the material. If None, then it is generated
+    data : DataContainer
+        data associated with this material
+
+    """
+    def __init__(self, uid=None, data=None, description=""):
```
tuopuu: I think that SSB expects that user can save a name parameter to the material with
```material_foo = Material(name="foo")```
(see #227)
Maybe there should be a name argument together with (optional) description. @ahashibon, could you please clarify this. In my opinion using the name would be sufficient in most cases, and the description could be used to add more detailed information about the material.

Comment (kemattil) [inline simphony/cuds/material.py]: Yes, I agree.

File: simphony/cuds/materials.py [hunk 8]
```diff
@@ -0,0 +1,99 @@
...
+
+        Raises
+        ------
+        ValueError:
+            If a material with this uid does not exist
+
+        """
+        self._materials[material.uid] = Material.from_material(material)
+
+    def get_material(self, uid):
```
tuopuu: A comment: I think get_material is a bit cumbersome to use with the uid. It would be easier for the user to make a search with the name parameter instead. Then it would be parallel with ```engine.get_dataset``` Cf.
```
engine.materials.get_material("steel")
# or
engine.materials.get_material(material_steel.uid)
```
If there's a good reason behind using uid here, then we can ignore this comment.

tuopuu:
@nathanfranklin, a few minor comments, and one bigger question with the name parameter for the Material class. Also, I would like to increase the CUDS file version number, because of the added column to the data container table class.

simphony/cuds/material.py
nathanfranklin: > I think that SSB expects that user can save a name parameter to the material with
material_foo = Material(name="foo")
(see #227)
Maybe there should be a name argument together with (optional) description

#227 had a mistake by adding the name.   that said, we could use 'name' instead of 'uid' for our unique definition.  it does seem easier for the user to use.

I am personally indifferent to either using a unique string 'name' or a 'uid' (but not both)  

(we wrote on the wiki that for MaterialRelation we would use a 'name' ..see faq at the end of https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/MaterialRelations_proposal)

nathanfranklin:
@tuopuu , @kemattil  thanks for the quick feedback.  i am out of the office today but will address the comments on monday

simphony/cuds/materials.py
nathanfranklin: Yes, they all have default values.

nathanfranklin:
@tuopuu , @kemattil : Besides the open "name" versus "uid" question, I have addressed your comments.  Please take a look.

nathanfranklin:
`Material` currently has an `uid` attribute (uuid.UUID) and this attribute is used to identify it in the `Materials` (`materials.get_material(uid)`).

@tuopuu , @kemattil and I are discussing if the `uid` should be replaced by an string attribute named `name`  (see comments in this PR. for example: https://github.com/simphony/simphony-common/pull/237#discussion_r46654698 )

Can someone else please provide their opinion on using a string `name` versus the UUID `uid` attribute?

simphony/cuds/materials.py
tuopuu: But what is the use of these attributes? The init method does not create these attributes to the ```Materials``` class.

nathanfranklin: @tuopuu , Sorry. I was looking at 'material.py' instead of 'materials.py'.  Cut and paste error :) Will fix it.

tuopuu: No problem, that's what I also thought it was. :)

nathanfranklin:
> Can someone else please provide their opinion on using a string name versus the UUID uid attribute? 

@ahashibon , @mehdisadeghi ,

tuopuu:
> Besides the open "name" versus "uid" question, I have addressed your comments.  Please take a look.

Otherwise looks good, :+1:

simphony/cuds/material.py
dpinte: @nathanfranklin the unique id notion needs to be general to the project. You can surely add a separate notion of name with all the required machinery (query_by_name, ensure name is unique, etc.) but it should not bypass the uid notion. If I am not wrong (and I may be ;-)), this is used in the serialization process and the link with other objects, no?

dpinte: You constructor could be  `def __init__(self, name, uid=None, data=None, desription='')` where `name` is mandatory but uid optional (and thus auto-generated)

nathanfranklin: >  If I am not wrong (and I may be ;-)), this is used in the serialization process and the link with other objects, no?

@dpinte  yes, that is correct.  

We use the unique id notion mostly (e.g. `"single_particle = particles.get_particle(particle.uid)`) but we do divert from using uid notion for the high-level datasets as they have a string `name` that is used to identify them (e.g. `foo_dataset = engine.get_dataset("foo")`).

nathanfranklin: > You can surely add a separate notion of name with all the required machinery (query_by_name, ensure name is unique, etc.) but it should not bypass the uid notion.

This is a good idea.  what do you think @tuopuu , @kemattil , @ahashibon

mehdisadeghi:
Does every material necessarily has a unique name? If yes it makes sense to replace `uid` with `name` or use the unique name as id. However, I'm afraid that we would be able to make any arbitrary material which would not necessarily have a proper name. Therefore I think it is better to keep the `uid` to make sure that every material in the system is uniquely identified. In order to have a human-friendly name I am in favor of having `name`, `description` or a list of `tags` for the material.

simphony/cuds/material.py
tuopuu: > You can surely add a separate notion of name with all the required machinery

Yeah, we should be able to query by name of by uid. There are at least three options to make it happen:
1. Write separate ```get_material_by_uid(uid)``` and ```get_material_by_name(name)``` methods
2. Write one method with two optional arguments ```get_material(name=None, uid=None)```
3. Write one method with one argument, and add a check inside the method, if this argument is of type ```str``` or ```uuid```, e.g.
```
class Materials(object):
     ...
     def get_material(id):
          if isinstance(id, uuid):
              ...
          elif isinstance(id, str):
              ...
```           
I prefer option 3. for it's simplicity.

ahashibon: Agree with @dpinte, adding name would be good.

ahashibon:
@mehdisadeghi: the description is an optional user defined text. It cannot be used to identify an entity in SimPhoNy uniquely. Name and UUID are better for this.

simphony/cuds/material.py
ahashibon: @tuopuu: it should be possible to explicitly choose a material based on uid OR name, I am in favour of the three method

simphony/cuds/abc_modeling_engine.py
ahashibon: for now it is ok to proceed this way.

simphony/cuds/material.py
nathanfranklin: >>    You can surely add a separate notion of name with all the required machinery

> Yeah, we should be able to query by name of by uid. There are at least three options to make it happen:
1. Write separate get_material_by_uid(uid) and get_material_by_name(name) methods
2. Write one method with two optional arguments get_material(name=None, uid=None)
3. Write one method with one argument, and add a check inside the method, if this argument is of type str or uuid, e.g.
```
class Materials(object):
     ...
     def get_material(id):
          if isinstance(id, uuid):
              ...
          elif isinstance(id, str):
              ...
```
I prefer option 3. for it's simplicity.

@ahashibon , @tuopuu , 

We should remember that the identification of Material (whether it be name-string, uid or both) is not just used in Materials which has been the focus of this discussion.  It is also to be used elswhere like identifying the material of low-level entities (`particle.data[CUBA.MATERIAL] = material_baz.uid`) and used in creating material relations (`material_relation_foo.materials = [material_bar.uid, material_baz.uid]`).    So I feel like it would be better just to use *primarily* a **single** identifier. 

We can also extend later whatever is holding the instances of 'Material' (currently 'Materials') so that they can be accessed easier.

So if we used 'uid' as this identifier (my current preference), we could  to have some query_by_name and other functions as suggested by @dpinte above and more ideas in the example by @tuopuu to whatever is maintaining the materials (currently 'Materials').  If we decide this approach, we can open another issue "Extend `Materials` to be able to query by `name`" and make this enhancement as another PR.

nathanfranklin:
@mehdisadeghi ,  see https://github.com/simphony/simphony-common/pull/237#discussion_r46973942

ahashibon:
go ahead with uuid, the name can be done later.

tuopuu:
> the name can be done later.

Agreed.

simphony/cuds/material.py
mehdisadeghi: I prefer an explicit method such as `get_material(uid)`. BTW, it is not clear for me why a `Material` class should have such a method at all, since it is the material itself, not a container. However, it makes sense for a container object of any form to have query methods such as `get`, `get_material`, or `get_material_by_name`. As for the parameters accepted by such methods, I am in favor of explicitly expecting `uid` and raising exception in case of incorrect argument.

nathanfranklin: > I prefer an explicit method such as get_material(uid). BTW, it is not clear for me why a Material class should have such a method at all, since it is the material itself, not a container. 

`get_material` is a method for `Materials` and not `Material`

nathanfranklin:
okay, is it okay to merge this then?

tuopuu:
> okay, is it okay to merge this then?

@nathanfranklin, go ahead. We can deal with the naming issue later.

kitchoi:
@nathanfranklin, just as I was testing integration with the visualisation plugins I found that the `dtype="uuid"` for Material is causing of error.  How about `dtype="numpy.object_"`?

kitchoi:
@nathanfranklin ah i see that the `dtype="uuid"` is used in `dummy_cuba_value` for generating dummy `uuid`.  That makes sense to me.  I will change the plugins then.

Issue #229, opened 2015-11-17
Add cuba_keys to add_dataset method of HDF5 CUDS


roigcarlo:
The add_dataset method of the hdf5 file wrapper (H5CUDS) should provide an optional cuba_keys argument to use for the supported cuba keywords that are needed to be copied to the file.

https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/Supported_CUBA_keys_for_container .

tuopuu:
In the current implementation of the H5CUDS backend, tables with 66 columns (corresponding to the current total of the CUBA keys) are used to store CUDS data regardless of the number of different CUBA keys defined for the saved objects. 

I understand that for solving this issue #229 we do not need to modify the backend, but instead just modify the add_particles, add_lattice and add_mesh so that unwanted CUBA keys are not stored in file (while there still exist all those 66 columns). The code itself could be something along the lines of 
```
# from h5_cuds.add_particles
...
if cuba_keys is not None:
        for p in particles.iter_particles():
                new_p = Particle.from_particle(p)
                new_p.data = new_p.data(cuba_keys[CUDSItem.PARTICLE])
                h5_particles.add_particles(new_p)
        ...
        else:
            h5_particles.add_particles(iter_particles())
            h5_particles.add_bonds(particles.iter_bonds())
```
Unfortunately, this might turn out to be highly inefficient. While there is currently no way of iterating over CUDSItems with masking a part of their CUBA keys, I don't see a better way to fix this issue #229. Any comments, @nathanfranklin, @roigcarlo, etc?

Anyway, a separate issue should be made for changing the file backend where not all columns are allocated.

nathanfranklin:
> Unfortunately, this might turn out to be highly inefficient. While there is currently no way of iterating over CUDSItems with masking a part of their CUBA keys, I don't see a better way to fix this issue
 
You mean, something like the following, right?  (I didn't follow the `new_p.data = new_p.data(cuba_keys[CUDSItem.PARTICLE])` line).
 
```
if cuba_keys is not None:
        for p in particles.iter_particles():
                p.data = DataContainer({key:data[key] for key in data if key in cuba_keys[CUDSItem.PARTICLE]})
                h5_particles.add_particles([p])
``` 

Yes, this seems to work.  Alternatively (as a future improvement), one could pass this `cuba_keys` information on to H5Particles's add_particles (or a future 'add_particles_filter' method).  

[...]

PR #244, opened 2015-12-03
PR: Feature choose which CUBA keys are stored in H5CUDS file

tuopuu:
This PR adds an optional cuba_keys argument to ```H5CUDS.add_dataset``` which can be used to select which CUBA keys are stored in the H5CUDS file. This PR addresses #229.


tuopuu:
There was some duplicate testing in test_h5_cuds for mesh and particles datasets, which is already covered in abc_check_engine template. I think we can remove these test.

nathanfranklin:
> There was some duplicate testing in test_h5_cuds for mesh and particles datasets, which is already covered in abc_check_engine template. I think we can remove these test.

It looks like there is lots of overlap with abc_check_engine but a few things appear to be be unique.  Like
for example https://github.com/simphony/simphony-common/blob/master/simphony/io/tests/test_h5_cuds.py#L90 , https://github.com/simphony/simphony-common/blob/master/simphony/io/tests/test_h5_cuds.py#L151-L154

Specifically, https://github.com/simphony/simphony-common/blob/master/simphony/io/tests/test_h5_cuds.py#L151-L154 where its the only place we test that we can't use the (i) H5Cuds file or (ii) any dataset proxy handles that were gotten from that file after the file been closed.  That said, it looks like this only performs the (ii) test for meshes.

tuopuu:
> (i) H5Cuds file or (ii) any dataset proxy handles that were gotten from that file after the file been closed.

I don't think the engine test template actually tests this either for any engine: are the handles usable after the engine has been closed/deleted? I think it could be added to the engine template and by that used everywhere.

tuopuu:
Reopening the file (even though it's kind of analogous to reconnecting/recreating the handle to a modeling engine) should be tested here, but for it we can write its own test.

nathanfranklin:
> are the handles usable after the engine has been closed/deleted?

We don't have 'close' in ABCModelingEngine

>>    (i) H5Cuds file or (ii) any dataset proxy handles that were gotten from that file after the file been closed.

> I don't think the engine test template actually tests this either for any engine: are the handles usable after the engine has been closed/deleted? I think it could be added to the engine template and by that used everywhere.

My initial feeling is that we should not alter the `abc_check_engine` for this PR.   `H5Cuds` has lots of overlap with ABCModelingEngine (all the 'XXX_dataset' methods) and so uses the testing framework for the engine (`abc_check_engine`) but H5Cuds is different beast.  And has some unique behavior (e.g. files can be closed). It also has some new differences to ABCModelingEngine that are added to this PR (i.e. 'add_dataset' has optional parameter) which need to be tested.

tuopuu:
> My initial feeling is that we should not alter the abc_check_engine

Yeah, it's not good make changes here. We'll have another issue for it if there's a need. I'll just clean the test_h5_cuds file a bit from the overlapping parts, and add the missing tests to the cuds-file handle related class. I'll make a push in a minute, so please take a look at it then. If there's something you'd still like to have there, let me know.

tuopuu:
Testing that closed file has the data is still missing, I'll add it.

nathanfranklin:
:+1:  Looks good to me.  If someone else has time, maybe someone else should take a look too before merging.

tuopuu:
Thanks @nathanfranklin, I think we can merge this. (No objections in a week.)

Issue #245, opened 2015-12-03
NaN value for PrimitiveCell.for_rhombohedral_lattice


kitchoi:
when the angle `alpha` gets close to 120 degree, the z coordinate of the third primitive vector gets vanishingly small.  Eventually `np.sqrt(sina**2 - ((cosa-cosa**2) / sina)**2)` [here](https://github.com/simphony/simphony-common/blob/master/simphony/cuds/primitive_cell.py#L196) is trying to compute the square-root of a negative value.

Edited:
Similar problem with `for_triclinic_lattice`
Should raise a ValueError when the quantity within sqrt is less than 0

Issue #240, opened 2015-11-30
lattice_tools module for checking lattice type


kitchoi:
Please take a look at this module and see if it would be appropriate to include it in simphony-common instead of simphony-mayavi

Code: https://github.com/simphony/simphony-mayavi/blob/master/simphony_mayavi/cuds/lattice_tools.py

The functions are implemented using the looser definitions of some lattices following the factory functions defined in simphony-common, e.g. cubic lattice is also a tetragonal/orthorhombic/rhombohedral/monoclinic lattice.  These details are documented.

tuopuu:
The code looks good, @kitchoi! From the lattice-Boltzmann perspective, it's not really essential to have it in simphony-common, but of course there might be other use-cases in SimPhoNy in future that would benefit from it. Therefore, my opinion is to add it here. However, I don't see a clear most favorite place for it to be added. There are a couple of options that come to mind:
1. Add it into the cuds/lattice.py file, but as they are not directly needed providing abc_lattice implementation, that might not the best place.
2. Create a new folder for these kind of tools, e.g. simphony/tools. People working with meshes might have needs for something similar: for example, calculating certain properties of a mesh. These tools could also be located in the same folder. Also, H5CUDS-file version conversion tools should be put in there as well.

kitchoi:
Thanks @tuopuu!
I also think that option 2 is better.

kemattil:
I also like option 2.

ahashibon:
This is a great set of utilities which we would also use for other cases, e.g., the crystal lattice builders. The best is to move them at your earliest convenience to simphony-common under tools or utils (choose what you like better, no difference here). I would also suggest a sub category as lattice utilities/tools, e.g. so user can do this: 

from simphony.utils import lattice_tools
from simphony.utils import mesh_tools 
from simphony.utils import crystal_lattice_tools as clt

kitchoi:
Thanks @ahashibon!  I have pushed a PR (#249) and it is ready for review; if it looks good then it can be merged.

PR #249, opened 2015-12-11
Add tools module and lattice_tools

kitchoi:
**Added lattice_tools as `simphony.tools.lattice_tools`**
     - please comment if you think a different naming or hierarchy is better)
**Added tests for the lattice_tools**
     - It takes a while because `hypothesis` tries to randomly construct many different primitive cells to catch edge cases; most of the time is spent on generating many primitive cell objects, not on calling `lattice_tools`
**Added `hypothesis` to dev_requirements.txt**
      - hugely useful for development (that's how I found out about #245)

Please play with it with your use cases.

Related to #240

### DIFF
--- a/simphony/tools/lattice_tools.py
+++ b/simphony/tools/lattice_tools.py
@@ -506,7 +506,8 @@ def is_base_centered_orthorhombic_lattice(p1, p2, p3):
     ''' Test if primitive vectors describe a base centered orthorhombic
     lattice
 
-    Also returns True for vectors describing a hexagonal lattice
+    Also returns True for vectors describing a cubic, tetragonal,
+    or hexagonal lattice
 
     Parameters
     ----------
@@ -517,16 +518,38 @@ def is_base_centered_orthorhombic_lattice(p1, p2, p3):
     -------
     output : bool
     '''
-    vec_lengths = map(vector_len, (p1, p2, p3))
-    factory = PrimitiveCell.for_base_centered_orthorhombic_lattice
+    # there are two types of primitive cells
+    # type 1: (a, 0, 0), (a/2, b/2, 0), (0, 0, c)
+    # type 2: (a/2, -b/2, 0), (a/2, b/2, 0), (0, 0, c)
 
-    for alpha, beta, gamma in permutations(vec_lengths, 3):
-        delta = 4.*beta**2.-alpha**2.
-        if (delta > 0. and
-                same_lattice_type(factory(alpha, numpy.sqrt(delta), gamma),
-                                  p1, p2, p3)):
+    dot_products = numpy.abs(map(numpy.dot, (p1, p2, p3), (p2, p3, p1)))
+    right_angles = dot_products < TOLERANCE
+    if right_angles.sum() < 2:
+        return False
+
+    lenA, lenB, lenC = map(vector_len, (p1, p2, p3))
+    equal_lengths = numpy.abs((lenA-lenB, lenB-lenC, lenC-lenA)) <= TOLERANCE
+
+    if equal_lengths.any() and all(right_angles):
+        return True
+    elif equal_lengths.any():
+        # find the pair of vectors that don't make a right angle
+        # if it is the type 2 cell, they should be equal length
+        if equal_lengths[numpy.where(~right_angles)[0][0]]:
             return True
-    return False
+
+    if not all(right_angles):
+        # try type 1
+        # find the two vectors that don't make right angles with the others
+        # the projection of one vector is half the another vector
+        not_right_angle = numpy.where(~right_angles)[0][0]
+        dot_product = dot_products[not_right_angle]
+        vec_mapping = {0: (p1, p2), 1: (p2, p3), 2: (p1, p3)}
+        v1, v2 = vec_mapping[not_right_angle]
+        return (numpy.abs(dot_product - numpy.dot(v1, v1)*0.5) < TOLERANCE or
+                numpy.abs(dot_product - numpy.dot(v2, v2)*0.5) < TOLERANCE)
+    else:
+        return False
 
 
 def is_monoclinic_lattice(p1, p2, p3):
--- a/simphony/tools/lattice_tools.py
+++ b/simphony/tools/lattice_tools.py
@@ -89,10 +89,11 @@ def same_primitive_cell_config(v1, v2, v3, p1, p2, p3, permute=True):
 
 
 def guess_primitive_vectors(points):
-    ''' Guess the primitive vectors underlying a given array of
-    lattice points (N, 3).  This assumes an ideal, fixed bravais
-    lattice with no defect within numerical errors, controlled by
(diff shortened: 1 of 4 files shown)


kitchoi:
Ready for review.

kitchoi:
Discussion with Adham led me to realise primitive cell configurations not included in the code.  Please hold off review for now.  Thanks!

kitchoi:
Added features:
- `lattice_tools.TOLERANCE` is added for controlling the numerical error tolerated
- `same_lattice_type` is renamed to `same_primitive_cell_config` as the different primitive cell configurations can be applied for the same lattice type
- use of `same_primitive_cell_config` in more symmetric lattice types are replaced by more explicit calculations for computational performance
- alternative primitive cell configurations not included in the PrimitiveCell module are supported for body-centered cubic, body-centered tetragonal, body-centered orthorhombic and base-centered orthorhombic (please review and see if I still miss some)
- `guess_primitive_vectors` is made more robust

Please review.  Thanks.

kitchoi:
A documentation (for sphinx) is not yet written.  I believe it is better to write a section for a set of tools under the tools module (lattice, crystal) once they are finalised.

simphony/tools/lattice_tools.py
```diff
@@ -0,0 +1,990 @@
...
+        tolerance for numerical errors
+
+    Returns
+    -------
+    match : bool
+    '''
+    targets = (v1, v2, v3)
+
+    # cosine angles between pairs of the target primitive vectors
+    target_cosines = numpy.abs([cosine_two_vectors(vec1, vec2)
```
tuopuu: @kitchoi, a question: Why numpy.abs() here (and L81)? This seems to interpret some "geometrically distinct" sets of vectors similar. I don't have an exact example in mind, but because cosine changes sign at pi/2, I think there might be a problem with some sets of vectors that have angles pi/2+c and pi/2-c, with c constant. Cosines of those angles would be the same. Maybe there's something I'm missing.

kitchoi: @tuopuu thanks for checking with care!  The absolute value here (and in the other functions) is to take care of the possibility of flipping a vector.  Two pairs of vectors with angles = (pi/2+c) and (pi/2-c) respectively are equivalent to flipping one of the two vectors by 180 degree.  These two sets of primitive vectors would generate the same lattice type.  An example of this is a hexagonal primitive cell with a 60 degree or 120 degree between the base vectors.

tuopuu: Thanks for the explanation, @kitchoi. It makes sense now to me. A 120 degree hexagonal lattice then equals a 60 degree one that is rotated 180 degrees around one of the main axises. Sorry, but I'll have to do something else now but I'll continue the review later this evening.

kitchoi: Thanks @tuopuu.  But now I can see that the docstring is leading to you think that the function should differentiate a hexagonal cell of 60 degree and 120 degree because I wrote there "geometrically similar".   I can imagine a user may have the same surprise when calling this function independently.  Maybe this is clearer?
~~~
Return True if a set of primitive vectors ``p1``, ``p2``,  ``p3`` 
describe a primitive cell that is geometrically similar to another
primitive cell described by (+/- ``v1``, +/- ``v2``, +/- ``v3``)
~~~
(I struggled to make it look pretty, hmm)

tuopuu: If the function only tests for angles and lengths, as it currently is, then the lattice type will be invariant under rotation, translation or reflection. A mathematical term for this collection of "geometrically similar" lattices is often called an "equivalence class". I think this is a really useful feature, because generally users don't often have the exact same configuration of the lattice that we have chosen as the representative of each lattice type. Currently SimPhoNy does not offer means to rotate or reflect the lattices, only translation is allowed by setting the origin parameter in the factory functions. 

Maybe we should change the docstrings once more to describe this behaviour.

tuopuu: Actually, translation invariance is not tested here as there is no comparison of the origin parameters. Primitive vectors themselves, of course, do not hold this information. So, please, ignore that part from my last comment.

Maybe the docstrings could say something along the lines of "Return True if a set of primitive vectors p1, p2, p3 describe a primitive cell that is similar under rotation and/or reflection to the cell described by v1, v2, v3."

simphony/tools/lattice_tools.py
```diff
@@ -0,0 +1,990 @@
...
+    p1, p2, p3: array_like
+        primitive vectors
+    tolerance: float
+        tolerance for numerical errors
+
+    Returns
+    -------
+    output : bool
+    '''
+    alpha2, beta2, gamma2 = map(numpy.dot, (p1, p2, p3), (p1, p2, p3))
```
tuopuu: A minor thing, but maybe these variables could be named differently as they do not represent angles here.

simphony/tools/lattice_tools.py
```diff
@@ -0,0 +1,990 @@
...
+        raise ValueError("Vectors contain invalid values")
+
+    if numpy.less_equal((lenA, lenB, lenC), tolerance).any():
+        raise ValueError("Edge lengths must be strictly positive")
+
+    cosines = numpy.abs(map(cosine_two_vectors,
+                            (p1, p2, p3), (p2, p3, p1)))
+
+    # base on loose definition: at least 2 angles are 90 degrees
+    return (cosines <= tolerance).sum() >= 2
```
tuopuu: I think we should also check here that none of the cosines is equal to 1 or -1. It would guarantee that the primitive cell is 3-dimensional (or has positive volume).

simphony/tools/lattice_tools.py
```diff
@@ -0,0 +1,990 @@
...
+        raise ValueError("Vectors contain invalid values")
+
+    if numpy.less_equal(vec_lengths, tolerance).any():
+        raise ValueError("Edge lengths must be strictly positive")
+
+    cosines = numpy.abs(map(cosine_two_vectors,
+                            (p1, p2, p3), (p2, p3, p1)))
+
+    for ivectors in permutations(range(3)):
+        alpha, beta, gamma = (vec_lengths[i] for i in ivectors)
```
tuopuu: I'd use some other naming than Greek letters here, as these variables refer to vector lengths and not angles.

tuopuu:
@kitchoi, I left a few minor comments between the lines. Nice looking code generally. This package will be very useful to crystallographers and lattice-Boltzmann enthusiasts. :+1:

kitchoi:
Thanks @tuopuu!  All fixed.  Please let me know if this is good to merge.  Thanks.

tuopuu:
@kitchoi, looks good. :+1:

I would have liked to still improve the docstrings in ```same_primitive_cell_config()```, but maybe it's not worth the time. :smile: It's not so easy to define what similarity there means without going too far into mathematics. Anyway, it should be intuitive enough what a comparison between cosines and length ratios here does.

PR #251, opened 2015-12-11
ValueError if the angle(s) for rhombohedral/triclinic are too big

### LINKED ISSUE #245: NaN value for PrimitiveCell.for_rhombohedral_lattice

kitchoi:
Fix #245

### DIFF
--- a/simphony/cuds/primitive_cell.py
+++ b/simphony/cuds/primitive_cell.py
@@ -189,11 +189,16 @@ class PrimitiveCell(object):
 
         cosa = np.cos(alpha)
         sina = np.sin(alpha)
+        delta2 = sina**2 - ((cosa-cosa**2) / sina)**2
+        if delta2 < 0:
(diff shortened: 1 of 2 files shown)


tuopuu:
@kitchoi, this looks good. I also checked if there are similar problems with the other lattice types but those seem to be fine. :+1:

kitchoi:
Great.  Thank you @tuopuu!

Issue #226, opened 2015-11-16
add MaterialRelation


nathanfranklin:
See https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/MaterialRelations_proposal

The MaterialRelation class has the following attributes:
    * name: a unique string name to identifying instances of these classes.
    * description: a user-defined description string
    * parameters: DataContainer containing the required parameters
    * supported_parameters: read-only attribute contains what CUBA values are required/allowed for the parameters
    * materials: list of uids of materials where this relation applies
    * kind: read-only attribute having a type CUBA which describes what kind of MaterialRelation this is (and subsequently what the supported_parameters are and what is the acceptable length(s) of the materials list). 

This requires: 
- [ ] Adding `MaterialRelation` class
- [ ] Adding tests and documentation
- [ ] Creating a `SP` object to contain these material relations (i.e. `add_material_relation`, `get_material_relation`, `delete_material_relation`, `update_material_relation`) (plus tests and documentation)

ahashibon:
Hold on with the SP!

nathanfranklin:
@roigcarlo , I see your branch where you started this work and there are some questions listed there. do you want to skype about this (i have time on Friday or Tuesday)?    Additionally, you can hold off on adding the material relations related to molecular dynamics (and related CUBA keywords) as I will cover those in #236  (LennardJones etc..)

roigcarlo:
Hi @nathanfranklin In fact I was planning to discuss that tomorrow during the meeting, or maybe later sure ( it shouldn't have been uploaded : P).
I'm not going to add any other material relation but I need a couple of classes in order to test that everything is working fine. 

Once I have some initial tests done I will make a PR and we can keep the discussion there with the code.

nathanfranklin:
Sounds good!

nathanfranklin:
Closed by #253

Issue #236, opened 2015-11-25
Create multiple MaterialRelations (i.e. LennardJones etc)


nathanfranklin:
This issue is related to #226 and https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/MaterialRelations_proposal  .  This issue to to add material relations related to MD (molectular dynamics)

[ ] extend CUBA to include keywords for each type of material relation and any needed parameters.  (Currently the known material relations are all MD ...LennardJones etc)
[ ] create a file containing description of various material relations (LennardJones etc)
[ ] write script to generate material relation classes from the description file.

nathanfranklin:
Closed by #253

PR #253, opened 2015-12-14
Feature material relation

roigcarlo:
This implements #226

The status of the PR is the following:
- [x] Add MaterialRelation class
- [x] Add Tests
- [x] Add documentation

As requested by @ahashibon the following point has not been implemented yet (and will be covered by #263).
-  Creating a SP object to contain these material relations (i.e. add_material_relation, get_material_relation, delete_material_relation, update_material_relation) (plus tests and documentation)


simphony/material_relations/material_relation.py
```diff
@@ -0,0 +1,118 @@
...
+        self._materials = materials
+        self._kind = kind
+
+    # Should we allow users to modify the id?
+    @property
+    def name(self):
+        return self._name
+
+    @name.setter
+    def name(self, value):
```
roigcarlo: I would like to discuss this.
In MaterialRelation the attirbute `name` has also the role of the `id`. In one and, I have my doubts that allowing the user to modify the id is correct, on the other hand, I can imagine that in some situations could be interesting to be able to change the name of the relation.

My proposal would be to use `uuid` instead of `name` as the default id, but I don't know if there is any particular reason for this election.

ahashibon: A CUDS (to avoid confusion: see below) will have many different material relations. The user needs a convinient way to check which relations are defined, and to be able to delete, update them etc. A name, as a tag that will allow the user to identify the material relation is therefore a must and is very useful. internally we can let every thing work with uuids if it is really going to make the code easier. But uuids are not user friendly (despite being the users best friend!)

CUDS: this is the new (original) CUDS object, that will be independent of an engine, and store all information relevant for one computational model..)

roigcarlo: @ahashibon Thanks for the clarification. Let's stick with the name then. ( In fact code is easier this way :+1:)

simphony/material_relations/coulomb.py
```diff
@@ -0,0 +1,56 @@
...
+    cutoffdistance : <type 'numpy.float64'>
+        Cutoff Distance
+    dielectriccontance : <type 'numpy.float64'>
+        Dielectric Contance
+
+    """
+
+    def __init__(
+        self,
+        name="Coulomb",
```
nathanfranklin: there should be no default value for the 'name' parameter.

simphony/scripts/material_realtion_generate.py
```diff
@@ -0,0 +1,244 @@
...
+        "class {MR_NAME}(MaterialRelation):\n".format(
+            MR_NAME=mr['class_name']
+        ),
+        "\n"
+    ]
+
+
+def generate_description_block(mr):
+    return [
+        "\t\"\"\" Automatically generated implementation of the\n",
```
nathanfranklin: Maybe stick the 'Automatically generated...' phrase not in the class's docstr but instead use it as a comment at the top generated python file.

simphony/core/cuba.yml
```diff
@@ -480,4 +480,46 @@
...
+  shape: [1]
+  type: double
+- description: Dielectric Contance
+  domain: []
+  key: DIELECTRIC_CONTANCE
+  name: DielectricContance
+  number: 72
+  shape: [1]
+  type: double
+- description: Lennard Jones
```
nathanfranklin: I think the two material relations (`LennardJones` and `Coulomb`) could be moved out of the CUBA list.  

Maybe it would be best to just use what is defined in `material_relations.yml` to create a seperate enum.  So maybe add 'number' to 'material_relations.yml' and generate a seperate enum (called CUBA_MATERIAL_RELATIONS or MATERIAL_RELATIONS?) 

Then the MaterialRelation attribute `kind` would be switched from `kind: CUBA` to something like `kind:CUBA_MATERIAL_RELATIONS.`

simphony/scripts/cuba_generate.py
```diff
@@ -84,7 +84,8 @@ def table(input, output):
         'uuid': 'String',
         'string': 'String',
         'double': 'Float64',
-        'integer': 'Int32'}
+        'integer': 'Int32',
```
nathanfranklin: The suggested change in https://github.com/simphony/simphony-common/pull/253/files#r47629563 would help remove the MaterialRelation-related stuff here.

simphony/core/cuba.yml
nathanfranklin: I think down the road maybe we could have a CUBA list which contained everything (including material relations). But currently we use the CUBA list as the list of everything that can go in a 'DataContainer'.   So until something like #29 happens, I think it would be best to have a separate enum for just the MaterialRelations (i.e. COULOMB and LENNARD_JONES)

simphony/material_relations/material_relation.py
```diff
@@ -0,0 +1,118 @@
...
+    parameters: DataContainer
+        the required parameters
+
+    supported_parameters: list of CUBA
+        CUBA values required/allowed for the parameters
+
+    materials: list of uids
+        materials where this relation applies
+
+    num_materials: list of int
```
nathanfranklin: I think 'num_materials' should be removed from both the __init__ parameter list as well as from this class attributes.  The number of materials can be derived from the 'kind'.

simphony/material_relations/material_relation.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Raises
+    ------
+    ValueError :
+        If the number of materials does not match with any value of
+        num_materials
+
+    """
+
+    def __init__(self, name, description, parameters, supported_parameters,
```
nathanfranklin: `supported_parameters` can be removed as a parameter as that information can be derived from `kind`

simphony/material_relations/coulomb.py
```diff
@@ -0,0 +1,56 @@
...
+            kind=CUBA.COULOMB
+        )
+
+    @property
+    def cutoff_distance(self):
+        return self.parameters[CUBA.CUTOFF_DISTANCE]
+
+    @cutoff_distance.setter
+    def cutoff_distance(self, value):
+        self.parameters[CUBA.CUTOFF_DISTANCE] = value
```
nathanfranklin: This needs to be reworked due to 'Material.parameters' being a property.

nathanfranklin:
@roigcarlo , I just left some comments.  Do you mind if I work on #236 or some of these comments on this branch (or in a separate branch and then have a PR back to this branch for you to review)?  I would plan on just working on it and handing it back to you by Wednesday afternoon or Thursday morning.

roigcarlo:
@nathanfranklin Thx for the review, I will start working on that right away. Of course feel free to upload changes directly in this branch. I think that #236 is related enough whit what we are doing here.

nathanfranklin:
> @nathanfranklin Thx for the review, I will start working on that right away.

If you are going to work on them right now, then I'll hold off then.

nathanfranklin:
>>  @nathanfranklin Thx for the review, I will start working on that right away.

>If you are going to work on them right now, then I'll hold off then.

@roigcarlo , or just let me know when you are done.

roigcarlo:
@nathanfranklin I don't expect to push anything until I leave the office, so if you want to work on this its fine

roigcarlo:
Well if you haven't started let me push the changes that I already have :)

simphony/material_relations/material_relation.py
```diff
@@ -0,0 +1,121 @@
...
+    def description(self):
+        return self._description
+
+    @description.setter
+    def description(self, value):
+        if isinstance(value, basestring):
+            self._description = value
+        else:
+            message = 'Expected string but received {!r}'
+            raise TypeError(message.format(type(basestring)))
```
kitchoi: `type(value)` instead of `type(basestring)`?

roigcarlo: Yup thx

simphony/material_relations/material_relation.py
```diff
@@ -0,0 +1,121 @@
...
+    def parameters(self):
+        return DataContainer(self._parameters)
+
+    @parameters.setter
+    def parameters(self, value):
+        self._check_parameters(value)
+        self._parameters = DataContainer(value)
+
+    def _check_parameters(self, value):
+        if (set(self._supported_parameters) != set(value.keys())):
```
kitchoi: do you mean 
~~~
if set(value.keys()) - set(self._supported_parameters):
    # check if any value.keys is not in supported_parameters
    raise...
~~~
my understanding is that the required parameters do not have to contain all the supported parameters, but i may be wrong.

roigcarlo: I think it is correct.

In the way this works updating parameters means reconstructing the DataContainer. if this only conatins a subset of the old values. then the rest will be lost.

For example consider a Coulumb material relations and the following code:

```python
mr = Coulumb(name='test_mr', materials=[mat1, mat2])

mr.cutoff_distance = 10
mr.dielectric_constant = 10

old_data = mr.properties
new_data = DataContainer()

new_data[CUBA.CUTOFF_DISTANCE] = 60

# At this point new_data contains:
# CUBA.CUTOFF_DISTANCE = 60.
# And nothing else, while old_data
# CUBA.CUTOFF_DISTANCE = 10
# CUBA.DIELECTRIC_CONSTANT = 10

mr.properties = new_data

# At this point we have deleted the dielectric_constant property.

```

If we check against all keys then we ensure that at least the user knows that he or she is modifying all the parameters of the relation.

Of course your point is interesting too, but as this behaviour is not defined elsewhere I probably say to stick with the most restrictive check.

kitchoi: Thanks.  I can see what you are trying to do.  Referring to https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/MaterialRelations_proposal, there is a "required" attribute for the cuba parameters and there is a differentiation between supported and required parameters.  Is this still going to be implemented or am I missing something?

kitchoi: I can see in the docstring that the `parameters` are required and the `supported_parameters` are required/allowed.  But the implementation seems to me that the `supported_parameters` are equivalent to the "required".  Again, did I miss something?

simphony/material_relations/material_relation.py
kitchoi: I agree with the point about user friendliness so there is a need for `name` to stay. As for using `name` to delete/update relations, it depends on how the relations are to be handled by the (new) CUDS, my opinion is that string is more prone to typo (and doesn't allow auto-completion by IDE...)

simphony/material_relations/material_relation.py
```diff
@@ -0,0 +1,121 @@
...
+        info = MATERIAL_RELATIONSHIP_INFORMATION[kind]
+        self._allowed_number_materials = info.allowed_number_materials
+        self._supported_parameters = [
+            parameter.cuba_key for parameter in info.supported_parameters]
+
+        self._name = name
+        self._kind = kind
+        self._description = description
+
+        self._check_materials(materials)
```
kitchoi: you could also do `self.materials = materials` and `self.parameters = parameters` so you reuse the setters.

simphony/material_relations/material_relation.py
roigcarlo: Yes, at this moment supported are required, I miss this as all of them are required in the examples. I will change that in the next push.

simphony/material_relations/material_relation.py
```diff
@@ -0,0 +1,121 @@
...
+        self._materials = materials
```
kitchoi: you could do `self.materials = materials` and `self.parameters = parameters` (if you rename the materials.setter `_set_materials` to `materials`) then you can reuse the setters which do the same thing.
Or are you trying to make materials a read-only attribute?

roigcarlo: @kitchoi Actually, materials had the ```@materials.setter``` as you suggest before merging with #257. I will change it back. @nathanfranklin if this was intentioned please tell us.

simphony/material_relations/material_relation.py
nathanfranklin: > Thanks. I can see what you are trying to do. Referring to https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/MaterialRelations_proposal, there is a "required" attribute for the cuba parameters and there is a differentiation between supported and required parameters. Is this still going to be implemented or am I missing something?

`supported_parameters`lists the CUBA keywords that should show up in the `parameters` DataContainer (i.e. exactly those CUBA should appear in `parameters`).  maybe `supported_parameters` should be renamed to `required_parameters`.

ahashibon: According to the original proposal, there is a slight difference between the supported and required. The required parameters are a subset of parameters that are supported by the MR. These are the parameters that the user must supply a value for upon creation of the instance of the MR. The rest of the (supported) parameters will get a default value if not specified. In other words, this is an attribute for keys that should tell when to raise if a parameter is not given a value when created.

ahashibon: Optional
remove required keyword. By default all supported parameters are required at initialisation, unless they have a default value in the markup, then this default value should be used in case the user does not specify anything else.

nathanfranklin: > Optional:
> remove required keyword. By default all supported parameters are required at initialisation, unless they have a default value in the markup, then this default value should be used in case the user does not specify anything else.

@ahashibon ,  i agree.  it seems like the best approach.  this is how its currently implemented in this PR.

nathanfranklin: maybe it would be best to remove `supported_parameters` as an attribute for the `Material` class as this list of allowed cuba can be derived from the attribute `kind` (through `material_relations.yml` -> `MATERIAL_RELATIONSHIP_INFORMATION` from `material_relations_information.py`)

ahashibon: most applications will be developed interactively, so a method to print out the supported parameters and a help is needed, whether this is done as supported parameters or as a help or doc method, will not matter.

simphony/material_relations/material_relation.py
nathanfranklin: :+1:  Fixed.

simphony/material_relations/material_relation.py
mehdisadeghi: @ahashibon We need to ask users to provide the absolute minimum data in their YAML file. As @nathanfranklin told, we can derive more information elsewhere. If a material relation has a CUBA key, it is well defined and user needs to either directly import that relation and instantiate it providing the required properties, or use a factory method, e.g. `create_relation('lennard_jones')`.

nathanfranklin: @ahashibon , @mehdisadeghi ,  if i understood you both correct what is decided is:

1) keep `supported_parameters` method

2) add factory method   `create_relation(CUDSMaterialRelation.LENNARD_JONES, name, materials)`. 

User can use the factory method:
```
lj_mat = create_relation(CUDSMaterialRelation.LENNARD_JONES, name, materials)
```

or the currently provided: 
```
from simphony.cuds.material_relations.lennard_jones import LennardJones
lj_mat = LennardJones(name, materials)
```

Is my understanding of those two points  correct?

roigcarlo: @nathanfranklin, @mehdisadeghi I agree that we can derive the information from let's say ```material_relations_information.py```, add get the default values from the CUBA keys, but the user will still have to supply such information If I understand the new approach correctly, right?

What I don't understand is, why we should add an additional factory method to do exactly the same as ```lj_mat = LennardJones(name, materials)```?

mehdisadeghi: When I suggested factory method, I was thinking of a class, responsible for parsing user provided YAML, and checking the SimPhoNy defined one, in order to create a properly configured MaterialRelation instance. However, the minimum required information  should be defined in the proposal.  Therefor, I think this should be discussed further on the proposal's wiki page.

nathanfranklin: > When I suggested factory method, I was thinking of a class, responsible for parsing user provided YAML, and checking the SimPhoNy defined one, in order to create a properly configured MaterialRelation instance. However, the minimum required information should be defined in the proposal. Therefor, I think this should be discussed further on the proposal's wiki page.

@mehdisadeghi ,  By "SimPhoNy defined one", you mean `material_relations.yml`, right?  We currently only have that single file to describe all SimPhoNy material relations.  

We use that file to generate the subclasses of `Material` (`lennard_jones.py` etc..), material relation kind enum (`cuds_material_relation.py`), class containing information about each kind of material relation (`material_relations_information.py`) as well as sphinx documentation.  

We also provide in the documentation (See "Extending SimPhoNy") information about so that SimPhoNy developers who want to add new material relations (by altering material_relations.yml) are aware of what files need to be generated.

I am not aware of any discussions about additional files to describe material relations.  Yeah, please propose that extension on the proposal's wiki page or on a new page.  We can complete this current PR based on the original/current proposal.

nathanfranklin:
@ahashibon , changed the name of the yml file

@roigcarlo , @kitchoi , @mehdisadeghi  @ anyone else,   I think this PR is ready to be merged 

(Note a container to hold the material relations will be done through a separate PR  issue #263 )

roigcarlo:
@nathanfranklin :+1: I will probably make some additional tests in a separate PR, but it should be safe to merge.

mehdisadeghi:
@nathanfranklin I think the new term was "definition" not "information", did I get it right?

nathanfranklin:
@mehdisadeghi , you are right.  Just fixed that.  See commit.  Any other comments (otherwise, I'll merge after the CI is done)

PR #255, opened 2015-12-16
Resize logo to 200px width. Fix #246.

### LINKED ISSUE #246: Readthedocs SimPhoNy logo appears too large
The logo on the readthedocs appears too large and ugly. 

snapshot1

RTD docs [states](http://sphinx-doc.org/config.html?highlight=html_logo#confval-html_logo) that a logo should be max 200px width:

>> html_logo
>> ... It is placed at the top of the sidebar; **its width should therefore not exceed 200 pixels**.

Therefore we have to resize the image to fix this issue. We can do it only for html docs or for everything. Let me know if there is anything else to consider.

mehdisadeghi:
I resized the simphony logo to comply with readthedocs guidelines. Please check the docs to see if everything remains OK.


mehdisadeghi:
Would somebody have a look if the docs build OK?

PR #257, opened 2015-12-17
Feature material relation update

nathanfranklin:
This makes some changes to #253 .

Primarily:

  *  some minor fixes
  *  adds unique enum for the material relations (see 52fc34d and its commit message)
  *  change some things related to testing
  *  documentation on generating things with material_relations_generate.py


nathanfranklin:
> @nathanfranklin I've looking the commits this morning but I didn't have time to finish with everything yet. You can merge this directly, I would have to do it in my branch anyway so its fine. If there is any issue I can change it directly there, so no problem.

@roigcarlo ,  perfect.

PR #259, opened 2016-01-04
PR: H5_Cuds (missing some documentation) [WIP]

tuopuu:
H5_Materials class for working with HDF5 backed Material data. H5CUDS has high level functions to saving and loading the whole Materials class to and from the file.


tuopuu:
@nathanfranklin, this PR is ready for review. I'll add the documentation once the general layout of the hdf5 backed Materials is fixed.

nathanfranklin:
@tuopuu ,  thanks.  I will take a look tomorrow.    

@mehdisadeghi , can you take a look too.

simphony/io/h5_materials.py
```diff
@@ -0,0 +1,194 @@
...
+    def _retrieve(self, row):
+        """ Return the Material from a table row instance.
+
+        """
+        cuba = self._position_to_cuba
+        mask = row['mask']
+        data = DataContainer({
+            cuba[key]: convert_from_file_type(row['data'][key], cuba[key])
+            for key, valid in enumerate(mask) if valid})
+        uid = uuid.UUID(hex=row['index'], version=4)
```
mehdisadeghi: It would be nice to break this into few lines to increase its readability:
masked = {}
for key, valid in enumerate(mask):
    if valid:
        masked[cuba[key]] = convert_from_file_type(row['data'][key], cuba[key])
data = DataContainer(masked)

simphony/io/tests/test_h5_cuds.py
```diff
@@ -255,5 +259,85 @@ def tearDown(self):
...
+        handle.close()
+        handle = self.file_factory()
+
+        with self.assertRaises(ValueError):
+            handle.add_materials(self.expected)
+
+        handle.close()
+        self.cleanup()
+
+
```
mehdisadeghi: At the moment the distinction between `test_h5_cuds` and `test_h5_*` is not very clear to me. Shouldn't this test goes into `test_h5_materials`? If it is in the right place, we might keep a convention of naming classes in this file as `Test----CudsOperations'.

tuopuu: Please notice that H5CUDS methods work with high level containers, particles, mesh, lattice and now Materials, while H5_xxx classes provide methods to work with atoms in these containers: particle, node, Material, etc... Although, I agree that there's probably some unwanted overlap on the tests.

simphony/io/tests/test_h5_materials.py
```diff
@@ -0,0 +1,230 @@
...
+from simphony.cuds.material import Material
+from simphony.io.h5_cuds import H5CUDS
+from simphony.io.h5_materials import H5Materials
+from simphony.core.cuba import CUBA
+from simphony.core.data_container import DataContainer
+from simphony.testing.utils import compare_materials
+
+
+class TestH5MaterialsAddGetMaterial(unittest.TestCase):
+
```
mehdisadeghi: This might be named `TestH5MaterialsMethods`, unless we are going to define multiple classes for testing groups of material methods. Just a suggestion though.

simphony/io/h5_materials.py
tuopuu: You're right, we can make it to look simpler. (This code was copied from ```DataContainerTable._retrieve```, but I'll leave that as is.)

simphony/io/tests/test_h5_cuds.py
```diff
@@ -255,5 +259,85 @@ def tearDown(self):
...
+        stored_materials = [material for material in stored.iter_materials()]
+
+        self.assertEqual(len(expected_materials_dict), len(stored_materials))
+        for material in stored_materials:
+            self.assertEqual(
+                material, expected_materials_dict[material.uid])
+
+        handle.close()
+        self.cleanup()
```
nathanfranklin: why not remove this line and move the work done in `cleanup` to `tearDown`

tuopuu: The idea was to recreate the file for each test, so that the Materials (group) would be empty at start of each of these tests. ```self.cleanup()``` just deletes this file, so that a new empty one is created in the next test.

simphony/io/tests/test_h5_materials.py
tuopuu: Thanks for the suggestion. I'll make the change.

simphony/io/tests/test_h5_cuds.py
tuopuu: I think, right now, ```add_materials``` etc. methods should look and behave similarly to ```add_dataset```etc. The difference with those two cases is that currently only one ```Materials``` container can be saved in a file. In future, ```Materials``` will be contained in the ```SD``` component, and we'll probably then have ```add_state_data``` method in ```H5CUDS```. I understand that in future it should be possible to save many "state datas" in a file, but only one ```Materials``` container per state data. This will be a big change to the file layout, and it will probably make ```add_materials``` redundant.

To make ```add_materials``` function similarly to ```add_dataset```, I could change it to _explicitly_ check if Materials table exists, and raise an error that asks user to first remove the existing container before adding a new one. Would this be OK? 

I can also change the naming of the test class to correspond with the others.

simphony/io/tests/test_h5_cuds.py
nathanfranklin: @tuopuu , yeah, if you put common cleanup steps (i.e. the code of `self.cleanup()`) in the `tearDown` you will have the behavior that you desire.  ```"The setUp() and tearDown() methods allow you to define instructions that will be executed before and after each test method."``` - https://docs.python.org/2/library/unittest.html)

nathanfranklin: Oops.  I see you already have `shutil.rmtree(self.temp_dir)` in the `tearDown`.  So `cleanup()` (i.e. `os.remove(self.filename)`) is redundant as you already remove the temp directory after each test.

You should just remove `self.cleanup()` and its uses.

simphony/io/tests/test_h5_cuds.py
nathanfranklin: >  In future, Materials will be contained in the SD component, and we'll probably then have add_state_data method in H5CUDS. I understand that in future it should be possible to save many "state datas" in a file, but only one Materials container per state data. This will be a big change to the file layout, and it will probably make add_materials redundant.

Yeah, I see your point.  We are waiting for the `SD` or some sort of `Cuds Model` (name?) but need to be able to save the materials now.  

What about removing `H5Materials` class (but reusing its code) and just have the following methods to get/set a single `Materials`:
```
class H5Cuds ...

def get_materials()  
    ....  # returns Materials
def set_materials(materials) 
    ...  # sets Materials (replaces materials)  
```
(or alternatively don't deal with 'Materials' and just have methods to get/delete/update/add 'Material')

simphony/io/tests/test_h5_cuds.py
tuopuu: Thanks for the tips, @nathanfranklin . I'll remove them.

simphony/io/tests/test_h5_cuds.py
tuopuu: If we put the code in ```H5Materials``` under H5CUDS class, then we should do the same with all other classes that pop up in future, like ```H5MaterialRelations```, ```H5BoundaryConditions```, etc. (although, they need not be necessarily named like that). I fear that it would make H5CUDS class cluttered if they all are put there. 

I prefer organizing the code in separate files, at least for now. Later these files can be combined under their respective higher level objects, such as ```H5StateData``` that would hold the code from ```H5Materials```, ```H5MaterialRelations``` etc. 

I'd still like to add that version control of the tables/structures is simpler with dedicated API for each data structure.

Maybe there's something in the big picture that I'm missing, so follow-up comments are very much appreciated.

nathanfranklin: >Maybe there's something in the big picture that I'm missing, so follow-up comments are very much appreciated.

Mostly, I feel that `MaterialRelations` is a different beast than the datasets. 

 For our datasets we defined abstract base classes (`ABCMesh`, `ABCParticles`, `ABCLattice`) and then have concrete implementations (e.g. `Particles`, `H5Particles`, `LammpsPaticles`...) that provide this "proxy" behavior (*"the a CUDS container is returned as a result of a ``get`` operation in a modelling engine the returned Container is a proxy to the information stored in the internal memory if the wrapper/modelling engine."*).  This allows for optimizations in each of the implementations for certain situations (e.g. a mesh had millions of points but user is only accessing a few hundred of them).

For `MaterialRelations`, we don't have such defined behavior as the class represents something much smaller (~ a few dozen materials).  So creating `H5Materials` seems unneeded.

My suggestion to remove `H5Materials` is just to simplify things in the short term.  I see your point about cluttering `H5Cuds` though.  Regardless, we shouldn't spend to much time thinking about it.  If others are happy with the current approach, lets just merge :)  

(if we do keep H5Materials, we need to rework the testing so that we don't repeat ourselves in the testing performed in `simphony/io/tests/test_h5_materials.py`` and 'simphony/cuds/tests/test_materials.py')

mehdisadeghi:
There will be a new PR regarding the YAML output.

PR #260, opened 2016-01-08
Add check for adding existing material

### LINKED ISSUE #258: Materials.add_material() does not raise a ValueError for already used uids
According to the documentation ```add_material()``` should raise a ```ValueError``` if a material with the same ```uid``` already exists in the ```Materials._materials``` dictionary. This is not happening and it's not tested either.

nathanfranklin:
Fixes #258

### DIFF
--- a/simphony/cuds/materials.py
+++ b/simphony/cuds/materials.py
@@ -24,10 +24,11 @@ class Materials(object):
             If a material with the same uid already exists
 
         """
-        try:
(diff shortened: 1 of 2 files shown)


nathanfranklin:
@tuopuu , @mehdisadeghi  can you take a look at fix

mehdisadeghi:
@nathanfranklin it looks allright.

PR #268, opened 2016-01-15
Add cuba keyword

khiltunen:
This PR adds CUBA keywords needed in CFD wrappers (openfoam, numerrin, kratos)


khiltunen:
@ahashibon are keyword names now acceptable ?

PR #272, opened 2016-01-19
set dtype of NUMBER_OF_TIME_STEPS to be Int

### LINKED ISSUE #271: CUBA.NUMBER_OF_TIME_STEPS is currently a float
Should `NUMBER_OF_TIME_STEPS` be int instead?
Current dtype for [`NUMBER_OF_TIME_STEPS` is float64](https://github.com/simphony/simphony-common/blob/master/simphony/core/keywords.py#L406)
Just been trying to run animations with kratos and mayavi and then got an error that the type is wrong.  Kratos expects `Int`.  The two should be consistent.

kitchoi:
Fix #271


nathanfranklin:
:+1:

PR #279, opened 2016-02-23
skip unnecessary hypothesis's health checks

kitchoi:
`hypothesis` changed FailedHealthChecks from warning to error.  I think it is legit to skip these tests.


mehdisadeghi:
Applying above changes, all tests pass and there is no sign of FailedHealthCheck.

PR #283, opened 2016-03-10
Adds a clone operation to the data container.

stefanoborini:
It includes the ability to extract a subset of the original data.
This PR is needed for issue simphony_mayavi#35.


stefanoborini:
I needed this PR for symphony mayavi, but I decided to use a different strategy. This performs a useless copy and creates a temporary object that is thrown away. I prefer to alter CUBADataAccumulator.append so that it limits the keys it internalises.

Rejecting

PR #284, opened 2016-03-11
[WIP] Count and expose available keys.

stefanoborini:
To solve simphony-mayavi/35, we need to know what keys are available in the
DataContainers of all the contained entities. Unfortunately, this information
is not available, and a linear scan is too expensive, so we need to do bookkeeping.


stefanoborini:
Rejected. After discussion with Kit I think there's a better way of doing this. We use VTKParticle as a data source for the info, which needs to scan the full data source anyway, therefore needs to accumulate and does not need to perform bookkeeping.

Issue #285, opened 2016-03-28
Query existing wrappers from each wrapper on initialization


mehdisadeghi:
Each wrapper should provide information about its wrapper classes. This issue is subject to change based on below proposal:

[https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/Wrapper_Registration](https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/Wrapper_Registration)

Issue #281, opened 2016-02-29
Modify current ABCModelingEngine to accept CUDS objects


roigcarlo:
Modify current ABCModelingEngine accordingly with the new schema.

Issue #277, opened 2016-02-08
Create the new CUDS computational model


roigcarlo:
The new computational model for CUDS needs to be implemented. 

The new description can be found in:
https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/images/e/e9/Status_of_the_SimPhoNy_framework_Specifications_D1.3_and_D1.6.pdf

Slides 27 to 29.

This will requeire to:
- [ ] Replace CM for SP
- [ ] Implement PE, which was part of the old SP
- [ ] Implement ME, which is partially implemented in #253
- [ ] Implement C ( Old BC with extended features)
- [ ] Implement M, which is already done in #237
- [ ] Rename SD for DS

mehdisadeghi:
@roigcarlo as we have partially implemented this, I would like to finish it. Shall we assign it to me and break it along the way into smaller tasks if necessary?

roigcarlo:
@mehdisadeghi In fact the intention was to discuss this during a meeting this week, if you don’t mind, but I can tentatively assign this to you if you want.

PR #286, opened 2016-03-28
Add CUDS support to SimPhoNy

mehdisadeghi:
In this PR we add a number of new features to SimPhoNy as part of #281 #285 #277 :

- [x] - Add CUDS class. The high level container for any CUDS component
- [x] - Modify engine base class to accept CUDS
- [x] - Add Simulation class. Think of it as a general wrapper
- [x] - Add base for engine extensions
- [x] - Load engine extenion metadata upon importing simphony
- [x] - Add query functions to get info about available engines
- [x] - Add store classes to hold information about StateData
- [x] - Add engine metadata tests
- [x] - Add simulation tests

### DIFF
--- a/doc/source/extending_simphony.rst
+++ b/doc/source/extending_simphony.rst
@@ -109,7 +109,7 @@ create a wrapper instance and return it.
 
 The other method is ``get_supported_engines``. This method must create and return a list of
 the engine's features. Features are simply a pair of physics equation and a list of methods
-that the corresponing egnine provides in order to solve the physics equation. Features are
+that the corresponing engine provides in order to solve the physics equation. Features are
 created by calling the ``create_engine_metadata_feature``. After creating features one must
 pass them to the ``create_engine_metadata`` method in order to create metadata objects. Each
 engine metadata object is a pair of engine name and a list of features provided by that engine.
--- a/doc/source/extending_simphony.rst
+++ b/doc/source/extending_simphony.rst
@@ -4,7 +4,7 @@ Extending SimPhoNy
 Plugins
 -------
 
-The SimPhoNy library can extended through two `entry points`_ for
+The SimPhoNy library can be extended through two `entry points`_ for
 contributing python modules that contain engine and visualisation components:
 
 - ``simphony.engine`` -- A python module that provides one or more
@@ -55,6 +55,70 @@ Will allow the user to import the new engine from inside the ``simphony`` module
 
 .. _entry points : http://pythonhosted.org/setuptools/pkg_resources.html#entry-points
 
+
+Plugin metadata
+---------------
+Each plugin should implement a subclass of ``simphony.engine.extension.ABCEngineExtension`` and
+put it inside its top-level package, i.e. the package that is registered at the entry point.
+
+This a sample from simphony-lammps-md plugin::
+
+    from simphony.engine import ABCEngineExtension
+    from simphony.engine import EngineInterface
+
+    class SimlammpsExtension(ABCEngineExtension):
+        def get_supported_engines(self):
+            lammps_features =\
+                self.create_engine_metadata_feature(MolecularDynamics(),
+                                                    [Verlet()])
+            liggghts_features =\
+                self.create_engine_metadata_feature(GranularDynamics(),
+                                                    [DEM()])
+            lammps = self.create_engine_metadata('LAMMPS',
+                                                 lammps_features,
+                                                 [EngineInterface.Internal,
+                                                  EngineInterface.FileIO])
+            liggghts = self.create_engine_metadata('LIGGGHTS',
+                                                   liggghts_features,
+                                                   [EngineInterface.FileIO])
+            return [lammps, liggghts]
+
+        def create_wrapper(self, cuds, engine_name, engine_interface):
+            use_internal_interface = False
+            if engine_interface == EngineInterface.Internal:
+                use_internal_interface = True
+            if engine_name == 'LAMMPS':
+                return LammpsWrapper(cuds=cuds,
+                                     engine_type=EngineType.MD,
+                                     use_internal_interface=use_internal_interface)
+            elif engine_name == 'LIGGGHTS':
+                return LammpsWrapper(cuds=cuds,
+                                     engine_type=EngineType.DEM,
+                                     use_internal_interface=use_internal_interface)
+            else:
+                raise Exception('Only LAMMPS and LIGGGHTS engines are supported. '
+                                'Unsupported eninge: %s', engine_name)
+
+
+SimPhoNy will inspect the plugins for subclasses of ``ABCEngineExtension``. SimPhoNy
+creates an instance of each class and keeps track of available wrappers and their capabilities.
+As implemented in the above example, there are two methods that should be implemented by
+wrapper developers. ``create_wrapper`` is a factory method that receives a ``CUDS``, an
+engine name and an interface (according to ``EngineInterface`` enum). This method must
+create a wrapper instance and return it.
+
+The other method is ``get_supported_engines``. This method must create and return a list of
+the engine's features. Features are simply a pair of physics equation and a list of methods
+that the corresponing egnine provides in order to solve the physics equation. Features are
+create by calling the ``create_engine_metadata_feature``. After creating features one must
+pass the to the ``create_engine_metadata`` method in order to create metadata objects. Each
+engine metadata object is a pair of engine name and a list of features provided by that engine.
+
+Having this class implemented in the wrapper plugin, one can query for the available engines::
+
+    >>> from simphony.engine import get_supported_engine_names
+    >>> get_supported_engine_names()
+    ['LIGGGHTS', 'LAMMPS']
 
 CUBA keywords
 -------------
--- a/doc/source/extending_simphony.rst
+++ b/doc/source/extending_simphony.rst
@@ -110,8 +110,8 @@ create a wrapper instance and return it.
 The other method is ``get_supported_engines``. This method must create and return a list of
 the engine's features. Features are simply a pair of physics equation and a list of methods
 that the corresponing egnine provides in order to solve the physics equation. Features are
-create by calling the ``create_engine_metadata_feature``. After creating features one must
-pass the to the ``create_engine_metadata`` method in order to create metadata objects. Each
+created by calling the ``create_engine_metadata_feature``. After creating features one must
+pass them to the ``create_engine_metadata`` method in order to create metadata objects. Each
 engine metadata object is a pair of engine name and a list of features provided by that engine.
 
 Having this class implemented in the wrapper plugin, one can query for the available engines::
--- a/simphony/cuds/abc_modeling_engine.py
+++ b/simphony/cuds/abc_modeling_engine.py
@@ -1,3 +1,6 @@
+"""This module is part of simphony-common package. It contains
+wrappers base class.
+"""
 from abc import ABCMeta, abstractmethod
 
 
@@ -21,6 +24,31 @@ class ABCModelingEngine(object):  # pragma: no cover
 
     """
     __metaclass__ = ABCMeta
+
+    def __init__(self, *args, **kwargs):
+        """Initialize the engine.
+
+        Parameters
+        ----------
+        cuds:
+            The CUDS computational model object
+
+        Returns
+        -------
+
+        """
+        self._cuds = kwargs.get('cuds')
+        self._load_cuds()
+
+    # This class is supposed to be overridden, however it is not labeled
+    # abstract, in order to be backward-compatible.
+    def _load_cuds(self):
+        """Load information from CUDS into the engine."""
+        pass
+
+    def get_cuds(self):
+        """Get current CUDS instance."""
+        return self._cuds
 
     @abstractmethod
     def run(self):
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -105,10 +105,3 @@ class CUDS(object):
         for component in self._store.iteritems():
             if isinstance(component, component_type):
                 yield component.name
-
-    def change_datastore(self, store):
-        """Replace the current dataset store with the given one.
-
-        This method is intented to be used from wrappers. A wrapper will
-        switch the store to its own proxy store after getting the cuds."""
-        self._dataset_store = store
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -33,11 +33,11 @@ class CUDS(object):
     @property
     def data(self):
         """Data container for CUDS items."""
-        return self._data
+        return DataContainer(self._data)
 
     @data.setter
-    def data_setter(self):
-        raise Exception('Replacing data container is not supported.')
+    def data_setter(self, value):
+        self._data = DataContainer(value)
 
     def add(self, component):
         """Adds a component to the CUDS computational model.
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -1,6 +1,7 @@
 """CUDS computational model implementation.
 
-This module contains the classes used to represent a computational model, based on SimPhoNy metadata.
+This module contains the classes used to represent a computational model,
+based on SimPhoNy metadata.
 """
 from ..core import DataContainer
 from .store import MemoryStateDataStore
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -51,10 +51,7 @@ class CUDS(object):
     @staticmethod
     def _is_dataset(obj):
         """Check if the object is a dataset."""
-        if isinstance(obj, (ABCParticles, ABCLattice, ABCMesh)):
-            return True
-        else:
-            return False
+        return isinstance(obj, (ABCParticles, ABCLattice, ABCMesh))
 
     @property
     def data(self):
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -66,7 +66,7 @@ class CUDS(object):
         self._data = DataContainer(value)
 
     def add(self, component):
-        """Adds a component to the CUDS computational model.
+        """Add a component to the CUDS computational model.
 
         Parameters
         ----------
@@ -95,8 +95,13 @@ class CUDS(object):
 
         Parameters
         ----------
-        component_id: str
+        component_id: uuid.UUID
             key of a CUDS component.
+
+        Returns
+        -------
+        component: CUDSComponent
+            a cuds component
         """
         if not isinstance(component_id, (str, uuid.UUID)):
             raise TypeError('ID should be of string or UUID type')
@@ -110,7 +115,7 @@ class CUDS(object):
 
         Parameters
         ----------
-        component_id: CUDSComponent uuid
+        component_id: uuid.UUID
             a component of CUDS, reflecting SimPhoNy metadata
         """
         if not isinstance(component_id, (str, uuid.UUID)):
@@ -135,9 +140,9 @@ class CUDS(object):
         component_type: CUDSComponent class
             a component of CUDS, reflecting SimPhoNy metadata
 
-        Returns
-        -------
-        An iterator over results
+        Yields
+        ------
+        An iterator over the components
         """
         for component in self._store.itervalues():
             if isinstance(component, component_type):
@@ -153,7 +158,8 @@ class CUDS(object):
 
         Returns
         -------
-        list: list of names of the items of the given type
+        names: list
+            names of the items of the given type
         """
         if any([issubclass(component_type, ABCParticles),
                issubclass(component_type, ABCLattice),
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -3,8 +3,13 @@
 This module contains the classes used to represent a computational model,
 based on SimPhoNy metadata.
 """
+import uuid
+
 from ..core import DataContainer
 from .store import MemoryStateDataStore
+from .abc_particles import ABCParticles
+from .abc_lattice import ABCLattice
+from .abc_mesh import ABCMesh
 
 
 # IDEA: add consistency check handlers/instances to the CUDS class (dynamic).
@@ -24,8 +29,15 @@ class CUDS(object):
         # Memory storage to keep CUDS components
         self._store = {}
 
-        # Data container
+        # The generic data container
         self._data = DataContainer()
+
+        # A map to find the object for a given id amoung datasets or
+        # simple components. Unfortunately, at the moment dataset
+        # containers do not have `uid' and therefore this workaround
+        # is necessary. This is a dict of key of any kind and a lambda
+        # that will return the corresponding value.
+        self._map = {}
 
     @staticmethod
     def _is_cuds_component(obj):
@@ -35,6 +47,14 @@ class CUDS(object):
         # TODO: replace with metadata type checks
         return all([hasattr(obj, 'name'),
                     hasattr(obj, 'data')])
+
+    @staticmethod
+    def _is_dataset(obj):
+        """Check if the object is a dataset."""
+        if isinstance(obj, (ABCParticles, ABCLattice, ABCMesh)):
+            return True
+        else:
+            return False
 
     @property
     def data(self):
@@ -57,19 +77,33 @@ class CUDS(object):
         if not self._is_cuds_component(component):
             raise ValueError('Not a CUDS component.')
 
-        # Store the component
-        # TODO: use uuid of the component - does not exist yet
-        self._store[id(component)] = component
+        # Add datasets separately
+        if self._is_dataset(component):
+            self._dataset_store.add(component)
+            # Datasets have a uuid field called uid
+            self._map[component.name] = \
+                lambda key=component.name: self._dataset_store.get(key)
+        else:
+            # Store the component. Any cuds item has uuid property.
+            component_id = str(component.uuid)
+            self._store[component_id] = component
+            self._map[component_id] = \
+                lambda key=component_id: self._store.get(key)
 
     def get(self, component_id):
         """Gets a component from the CUDS computational model.
 
         Parameters
         ----------
-        component_id: CUDSComponent's UUID
-            uuid of a CUDS component.
+        component_id: str
+            key of a CUDS component.
         """
-        return self._store.get(component_id)
+        if not isinstance(component_id, (str, uuid.UUID)):
+            raise TypeError('ID should be of string or UUID type')
+        if isinstance(component_id, uuid.UUID):
+            component_id = str(component_id)
+        if component_id in self._map:
+            return self._map[component_id]()
 
     def remove(self, component_id):
         """Removes a component from the CUDS computational model.
@@ -79,7 +113,18 @@ class CUDS(object):
         component_id: CUDSComponent uuid
             a component of CUDS, reflecting SimPhoNy metadata
         """
-        self._store.remove(component_id)
+        if not isinstance(component_id, (str, uuid.UUID)):
+            raise TypeError('ID should be of string or UUID type')
+        if isinstance(component_id, uuid.UUID):
+            component_id = str(component_id)
+
+        component = self.get(component_id)
+        if not component:
+            raise KeyError('%s' % component_id)
+        if self._is_dataset(component):
+            self._dataset_store.remove(component.name)
+        else:
+            del self._store[component_id]
 
     # This should be query method
     def iter(self, component_type):
@@ -89,6 +134,10 @@ class CUDS(object):
         ----------
         component_type: CUDSComponent class
             a component of CUDS, reflecting SimPhoNy metadata
+
+        Returns
+        -------
+        An iterator over results
         """
         for component in self._store.itervalues():
             if isinstance(component, component_type):
@@ -101,7 +150,15 @@ class CUDS(object):
         ----------
         component_type: CUDSComponent class
             a component of CUDS, reflecting SimPhoNy metadata
+
+        Returns
+        -------
+        list: list of names of the items of the given type
         """
-        for component in self._store.iteritems():
-            if isinstance(component, component_type):
-                yield component.name
+        if any([issubclass(component_type, ABCParticles),
+               issubclass(component_type, ABCLattice),
+               issubclass(component_type, ABCMesh)]):
+            return self._dataset_store.get_names()
+        else:
+            return [cp.name for cp in self._store.itervalues()
+                    if isinstance(cp, component_type)]
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -32,7 +32,7 @@ class CUDS(object):
         # The generic data container
         self._data = DataContainer()
 
-        # A map to find the object for a given id amoung datasets or
+        # A map to find the object for a given id among datasets or
         # simple components. Unfortunately, at the moment dataset
         # containers do not have `uid' and therefore this workaround
         # is necessary. This is a dict of key of any kind and a lambda
@@ -96,7 +96,7 @@ class CUDS(object):
         Parameters
         ----------
         component_id: uuid.UUID
-            key of a CUDS component.
+            key of a CUDS component
 
         Returns
         -------
@@ -142,7 +142,7 @@ class CUDS(object):
 
         Yields
         ------
-        An iterator over the components
+        iterator over the components of the given type
         """
         for component in self._store.itervalues():
             if isinstance(component, component_type):
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -1,0 +1,107 @@
+"""CUDS computational model implementation.
+
+This module contains the classes used to represent a computational model, based on SimPhoNy metadata.
+"""
+from ..core import DataContainer
+from .store import MemoryStateDataStore
+
+
+# IDEA: add consistency check handlers/instances to the CUDS class (dynamic).
+class CUDS(object):
+    """CUDS computational model."""
+    def __init__(self):
+
+        # Add datasets to memory datastore by default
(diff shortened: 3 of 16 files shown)


simphony/cuds/simulation.py
```diff
@@ -0,0 +1,42 @@
...
+        Parameters
+        ----------
+        cuds: CUDS
+            a cuds object containing simulation data
+        wrapper_name: str
+            A SimPhoNy wrapper name, should be of of registered engines.
+        """
+        wrapper = get_wrapper(wrapper_name)
+        if not wrapper:
+            raise Exception('Unknown wrapper: %s' % wrapper_name)
```
kitchoi: It would be helpful to print the list of known `wrapper_name` in the exception message.

simphony/cuds/abc_modeling_engine.py
```diff
@@ -22,6 +25,31 @@ class ABCModelingEngine(object):  # pragma: no cover
...
+
+        Returns
+        -------
+
+        """
+        self._cuds = kwargs.get('cuds')
+        self._load_cuds()
+
+    # This class is supposed to be overridden, however it is not labeled
+    # abstract, in order to be backward-compatible.
```
kitchoi: It might just be me: sometimes it is good to not be backward-compatible and force changes that are required.  And it seems to me this is the case here.

simphony/engine/__init__.py
```diff
@@ -7,6 +7,25 @@
...
+    Parameters
+    ----------
+    wrapper_name: str
+        name that wrapper is registered with.
+    """
+    return _WRAPPER_REGISTRY.get(wrapper_name)
+
+
+def get_wrappers():
+    """Return a list of know wrappers."""
```
kitchoi: typo: "known"

simphony/cuds/simulation.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Parameters
+    ----------
+    cuds: CUDS
+        A cuds object which contains model information.
+    wrapper_name: str
+        Name of the wrapper to launch the simulation with.
+    """
+    def __init__(self, cuds, wrapper_name):
+        self._cuds = cuds
```
kitchoi: I think the Simulation wrapper is supposed to hold multiple cuds objects.  (referring to https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/File:Status_of_the_SimPhoNy_framework_Specifications_D1.3_and_D1.6.pdf)
According to the slides, I imagine that "self._cuds" should be a sequence of cuds objects, and after each run, the Simulation wrapper may append to "_cuds" with the output of the run.

simphony/cuds/simulation.py
mehdisadeghi: Thanks, I will apply that.

simphony/cuds/abc_modeling_engine.py
mehdisadeghi: I agree. This is temporary, however. As soon as the changes are finalized, we make it abstract.

kitchoi: Sounds good to me :)

simphony/cuds/simulation.py
mehdisadeghi: This has been subject of lots of discussions here. Shall we have a cuds object that keeps track of its state changes or shall we have a separate bookkeeper object? Soon I will make changes to support multiple cuds. Consider this one as the initial state, i.e. time_step=0.

simphony/cuds/abc_modeling_engine.py
```diff
@@ -22,6 +25,31 @@ class ABCModelingEngine(object):  # pragma: no cover
     """
     __metaclass__ = ABCMeta
 
+    def __init__(self, *args, **kwargs):
```
kitchoi: Is the flexibility in the __init__ a concern for interoperability?  (think the "use_internal_interface" switch in lammps)

simphony/cuds/model.py
```diff
@@ -0,0 +1,107 @@
...
+                    hasattr(obj, 'data')])
+
+    @property
+    def data(self):
+        """Data container for CUDS items."""
+        return self._data
+
+    @data.setter
+    def data_setter(self):
+        raise Exception('Replacing data container is not supported.')
```
SGGgarcia: Shouldn't we have the same the behaviour as we do with Mesh, Lattice and Particles ítems? Returning a copy of the data, let the user work with the copy, and when he is happy with it set the new data as a copy (screenshot model)

simphony/cuds/model.py
```diff
@@ -0,0 +1,107 @@
...
+        # When I see a bird that walks like a duck and swims like a duck
+        # and quacks like a duck, I call that bird a duck
+        # TODO: replace with metadata type checks
+        return all([hasattr(obj, 'name'),
+                    hasattr(obj, 'data')])
+
+    @property
+    def data(self):
+        """Data container for CUDS items."""
+        return self._data
```
SGGgarcia: See my comment below :+1:

simphony/cuds/abc_modeling_engine.py
mehdisadeghi: This is a good point indeed. If this information is also part of CUDS then there is no need for this. However, the wrappers are supposed to be instantiated by Simulation object and this details are hidden from outside users.

simphony/cuds/model.py
mehdisadeghi: Good catch, I will change this accordingly.

simphony/cuds/abc_modeling_engine.py
kitchoi: > the wrappers are supposed to be instantiated by Simulation object and this details are hidden from outside users.

Good point.  In addition, even without putting the arguments into CUDS, we would still have the issue with the docstring of `Simulation.__init__`:  A user new to the framework will need to use `get_wrapper(wrapper_name)` in order to fetch the class and its docstring.

Alternatively, this may be explicitly as for what the user should do:
```
cuds = CUDS()
sim = Simulation(cuds, LammpsWrapper(use_internal_interface=True))
```
Note: my understanding is that we should restrict a new engine wrapper instance being created for each individual `Simulation` instance in case it shares the same cuds with another `Simulation`, which could be a problem depending on how the engine wrapper is implemented (e.g. a file-io wrapper maybe initialised to look at a particularly file directory and different Simulation sharing the same wrapper ends on overwriting the files).

simphony/engine/__init__.py
mehdisadeghi: Thanks, fixed that.

simphony/cuds/abc_modeling_engine.py
kitchoi: I believe one of the motivations for having `Simulation` to instantiate an modeling engine wrapper internally is because some data from the cuds is required for the instantiation.  This is the part I don't know much about: What are these metadata that are required for instantiation?  Is "use_internal_interface" the only one?  Why can't we ask the user to initialise the modeling engine wrapper themselves?
(The second motivation is for documenting the simulation: what engine is used; in which case `Simulation` could save `repr(modeling_engine)` to somewhere.  And the `__repr__` method of the modeling engine can be overloaded to include more information, which is up to the dev team of the engine)

mehdisadeghi: > A user new to the framework will need to use get_wrapper(wrapper_name) in order to fetch the class and its docstring

This will be unnecessary. Users would need to fill a CUDS using the generated metadata classes and create and run a simulation with it.

> wrapper instance is created for each individual Simulation

Even though we currently create a new instance per simulation, we have no means of controlling whether multiple simulations have consistent configuration. To make it happen, we would need new parts in place. We can address some of these problems in client/server subsystem.

> What are these metadata that are required for instantiation?

The motivation behind instantiating wrappers internally, is to have CUDS object which contain all the necessary information about a computational model to start a simulation. Information such as engine name, parallelism, being internal or fileio, etc. This way, Simulation will have enough information to create an instance of a suitable wrapper (which is defined in the CUDS), configure it and then run it.

The other goal here, is to hide the complexity of the wrappers from users and move it to SimPhoNy side. This will help us to move things around internally without affecting the way SimPhoNy is being used.

kitchoi: > In addition, even without putting the arguments into CUDS, we would still have the issue with the docstring of Simulation.__init__: A user new to the framework will need to use get_wrapper(wrapper_name) in order to fetch the class and its docstring.

My above comment regarding fetching the docstring is referring to the situation when `*args` and `**kwargs` are allowed in `Simulation.__init__` (otherwise I don't know how to comply with `ABCModelingEngine.__init__(self, *args, **kwargs)`.  If all the required parameters (e.g. use_internal_interface) are to be moved to CUDS, then `ABCModelingEngine.__init__(self, *args, **kwargs)` would have been `ABCModelingEngine.__init__(self)` and `Simulation.__init__` accepts no more `*args` or `**kwargs` (as it is now).  Have I understood the intention correctly?

mehdisadeghi:
@kitchoi @SGGgarcia @roigcarlo @tuopuu Please review the latest changes to this PR.

simphony/cuds/model.py
```diff
@@ -0,0 +1,108 @@
...
+This module contains the classes used to represent a computational model,
+based on SimPhoNy metadata.
+"""
+from ..core import DataContainer
+from .store import MemoryStateDataStore
+
+
+# IDEA: add consistency check handlers/instances to the CUDS class (dynamic).
+class CUDS(object):
+    """CUDS computational model."""
```
kitchoi: Good place here to spell the long form of "CUDS"?

simphony/cuds/model.py
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self):
+
+        # Add datasets to memory datastore by default
+        self._dataset_store = MemoryStateDataStore()
+
+        # Memory storage to keep CUDS components
+        self._store = {}
+
+        # Data container
+        self._data = DataContainer()
```
kitchoi: What does CUDS.data hold? Information that does not belong to any of the CUDS Components?  The scope seems broad.

simphony/cuds/store.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class MemoryStateDataStore(ABCStateDataStore):
+    """A dataset store that keeps everything in memory."""
+    def __init__(self, *args, **kwargs):
+        self._datasets = {}
+
+    def add(self, dataset, *args, **kwargs):
+        """Add the dataset to the store."""
+        self._datasets[dataset.name] = dataset
```
kitchoi: check if `dataset.name` already exists?  because otherwise it would get overwritten.

simphony/cuds/store.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def iter_datasets(self, names=None):
+        """Iterate over all or a subset of datasets."""
+        if not names:
+            names = self._datasets.keys()
+
+        for key in self._datasets:
+            if key in names:
+                yield self._datasets.get(key)
+
```
kitchoi:
```
for key in names:
    yield self._datasets(key)
```
or
```
for key in names:
    yield self._datasets[key]
```
The former yields None if the key does not exist, the latter raises a KeyError.  Depends on which one makes more sense here (which I believe it the latter)
Same goes for the get function https://github.com/simphony/simphony-common/blob/feature-cuds/simphony/cuds/store.py#L44

simphony/cuds/simulation.py
```diff
@@ -0,0 +1,28 @@
+"""Contains simulation controllers."""
+from . import CUDS
```
kitchoi: `from .model import CUDS`?

simphony/cuds/model.py
mehdisadeghi: Right, I added more docs there.

simphony/cuds/simulation.py
mehdisadeghi: Having `.` import it from `__init__` module which I think is slightly better, but I have no strong opinion about it.

simphony/engine/extension.py
```diff
@@ -0,0 +1,204 @@
...
+            return extension.create_wrapper(cuds,
+                                            engine_name,
+                                            engine_interface)
+        else:
+            raise EngineManagerException(
+                'Invalid engine name: %s' % engine_name)
+
+    def get_supported_engines(self):
+        """Show a list of supported engines."""
+        return self._engine_extensions.keys()
```
kitchoi: `ABCEngineExtension.get_supported_engines` returns "list: a list of EngineMetadata objects"
Might be good to keep it that way here too?

simphony/engine/extension.py
```diff
@@ -0,0 +1,204 @@
...
+            The interface to the engine, internal or fileio.
+        """
+        if engine_name in self._engine_extensions:
+            extension = self._engine_extensions[engine_name]
+            return extension.create_wrapper(cuds,
+                                            engine_name,
+                                            engine_interface)
+        else:
+            raise EngineManagerException(
+                'Invalid engine name: %s' % engine_name)
```
kitchoi: Might be good to include contents of `self._engine_extensions.keys()` in the error message.

simphony/engine/extension.py
```diff
@@ -0,0 +1,204 @@
...
+                'The provided object is not a module: %s' % plugin)
+
+        for name, value in inspect.getmembers(plugin, inspect.isclass):
+            if not inspect.isabstract(value) and\
+                    issubclass(value, ABCEngineExtension):
+                # Load the corresponding engine extension
+                extension = value()
+                if not isinstance(extension, ABCEngineExtension):
+                    raise ValueError('Expected ABCEngineExtension, got %s' %
+                                     extension)
```
kitchoi: `issubclass(value, ABCEngineExtension)` is already done so the `isinstance` check here is not needed.

kitchoi:
Great work @mehdisadeghi !  Look forward to the final implementation of the ProxyStateDataStore/MemoryStateDataStore/change_datastore.  The rest looks fine.
I have a few minor comments above and the following:
```
class ABCEngineExtension(object):
    @abstractmethod
    def get_supported_engines(self):

     @abstractmethod
     def create_wrapper(self, ...):
```
Should they be (abstract) classmethod and staticmethod instead?  In which case, https://github.com/simphony/simphony-common/blob/feature-cuds/simphony/engine/extension.py#L154:
```
        for name, value in inspect.getmembers(plugin, inspect.isclass):
            if not inspect.isabstract(value) and\
                    issubclass(value, ABCEngineExtension):
                # Load the corresponding engine extension
                extension = value()
                if not isinstance(extension, ABCEngineExtension):
                    raise ValueError('Expected ABCEngineExtension, got %s' %
                                     extension)
                self.add_extension(extension)
```
could become
```
        for name, extension in inspect.getmembers(plugin, inspect.isclass):
            if issubclass(extension, ABCEngineExtension) and extension is not ABCEngineExtension:
                # Load the corresponding engine extension
                self.add_extension(extension)
```
so you don't have to instantiate the extension.

It seems that `EngineManager` is meant to be used as a Singleton (correct me if this is not the case).  Would it be possible to have its methods in a module instead of a class?

simphony/cuds/model.py
mehdisadeghi: I have to rethink this part. The initial reason behind it is, well, everybody likes a `data` attribute! Apart from that, it is there to comply with CUDS metadata model. However, I see some inconsistencies here. Maybe the data attribute should reflect the items inside the internal store, but I have to see if it is feasible at all.

simphony/cuds/store.py
mehdisadeghi: Good catch! I added a check for it.

simphony/cuds/store.py
mehdisadeghi: Since `key` is already read from the `_dataset` dictionary itself, it will always have a value and no KeyError will raise. However I changed it, in order to stay on the safer side :)

For the `get_method` method, it is intentional, similar to python's `dict.get` method.

simphony/engine/extension.py
mehdisadeghi: This method is intended to be a query function for user to get informed about available engines. I renamed this one to `get_supported_engine_names` and added a new one called `get_supported_engines` which returns a list of EngineMetadata as you mentioned earlier.

simphony/engine/extension.py
mehdisadeghi: More informative error messages are welcome!

simphony/engine/extension.py
mehdisadeghi: `isinstance` checks the result of invoking `value()` which might be of different type from `value` itself, for any reason. One example is overring __new__ method. I know, I am a little paranoid here!

kitchoi: That's true... that would be an evil move though 😈

mehdisadeghi:
> Should they be (abstract) classmethod and staticmethod instead?

A good question indeed. This is possible of course. However, python has no builtin notion of abstract class or static method. The choice of method/class based approach vs function/module based approach to implement extensions is more or less a matter of taste. Moreover, it represents two different programming paradigms: object oriented programming vs functional programming style. In this case, I think the former lets us to build a more flexible framework. Turning those methods into functions reduces the benefits of having an extension class all together. One can also argue that, we can replace classes (as in your second question), with modules. A module is an inappropriate choice when we have classes right at hand in Python.

> Would it be possible to have its (EngineManager) methods in a module instead of a class?

Please refer to the above answer. 

@kitchoi thanks for your very good and detailed reviews. Feel free to [continue](https://simphony.slack.com/archives/common) the discussion.

kitchoi:
@mehdisadeghi Thanks for your explanation.

As for my question about using a module instead, I was only referring to `EngineManager` which is used internally and with one instance.  If you meant for it to be subclassed or have multiple instance of such class, my comment would be irrelevant and please ignore.  I think module is essentially the implementation of a Singleton, at least in Python.  So if you are certain of the Singleton design, go with the module, the code would be cleaner and easier to maintain.  But at the end of the day, the implementation of `EngineManager` is very much internal that it is possible to modify it without upsetting the API; I am fine with it going ahead, so that we don't obstruct the next stage of development.

> However, python has no builtin notion of abstract class or static method.

True and false.  In Python 2, people have been making their own implementation for abstract class and abstract static methods, so much that there are now builtins in Python 3 :)

>  Turning those methods into functions reduces the benefits of having an extension class all together.

Yes I agree, and hence `ABCEngineExtension` should be a class as it is meant to be subclassed, I have no doubt about that.   The questions for whether to restrict `get_supported_engines` and `create_wrapper` as classmethod/staticmethod stemmed from my observation of (1) how the extension is implemented in `simlammps`, in which `create_wrapper` does not touch `self` and `get_supported_engines` only uses `self.create_wrapper`, plus (2) the `EngineManager` assumes that the instantiation of the extension requires no more specification (`extension = value()`).  So would we expect these methods to ever need to access the instance `self`?  If not, wouldn't it be clearer to restrict them as abstract classmethod (staticmethod might be too strict, actually).  Note that this detail affects the public API, so once we settle on it, we probably don't want to change it in the future.

mehdisadeghi:
> If you meant for it to be subclassed or have multiple instance of such class

Neither of these. It is supposed to be replace dynamically if necessary, and to be part of a larger framework (which does not exist yet).

> module is essentially the implementation of a Singleton

This can be true. However, one might end up importing different modules or passing module instances around in order to assign another manager for handling extensions. I am more comfortable with classes (or biased toward them lets say).

> there are now builtins in Python 3

Good to know that!

> would we expect these methods to ever need to access the instance self

Yes. One example use case would be adding hooks before/after creating wrappers. The other one would be calling super class implementation of the method (which is empty now).

kitchoi:
>  It [EngineManager] is supposed to be replace dynamically if necessary, and to be part of a larger framework (which does not exist yet). ...
> However, one might end up importing different modules or passing module instances around in order to assign another manager for handling extensions. 

Do you mean modifying `simphony.engine._ENGINE_MANAGER` dynamically?  Then perhaps it should be a public variable? (i.e. `simphony.engine.engine_manager`).  If that's the case, I can see why a class would be better.  This usage should be documented, though.

> One example use case would be adding hooks before/after creating wrappers.

I would much prefer no side effect for a function that returns something.  Why is this needed?

> The other one would be calling super class implementation of the method (which is empty now).

I think you can pass `type` as the second argument for `super`.

mehdisadeghi:
> Do you mean modifying simphony.engine._ENGINE_MANAGER dynamically?

No, not at all. That variable is just a workaround for missing platform pieces. If we put a component manager somewhere in simphony, this variable would be filled internally using a service call to a service from an underlying package.

> I would much prefer no side effect for a function that returns something

I agree that side-effects are evil, and we don't want them. My main purpose here is making system more dynamic, by putting place-holders for possible future extensions. In this case, supporting an event-system or registering hook callables which should be called before/after wrapper creation.

kitchoi:
Thanks @mehdisadeghi for your patience!   I see lots are planned (services/hooks) and I don't yet know them, so please bear with me.  Your comments help me to get a better idea on these plans.

>  If we put a component manager somewhere in simphony, this variable would be filled internally using a service call to a service from an underlying package. [update: this variable will be replaced with corresponding service call all together]

Without knowing what these services are about, I cannot comment further on the `EngineManager` class.

As for the implementation of the methods in ABCEngineExtension, I guess the problem for me is that I don't yet see why we would expect multiple instances of an EngineExtension, hence the "access to instance" question.  It would be very helpful (for this PR and future ones) if I can get further insight on this.  Thanks.

mehdisadeghi:
Since this PR is growing in size, I prefer to merge this and continue working on the other issues in a separated PR. Having this PR merged, others can use the lammps-md example as a reference implementation and implement the wrapper extension part.

simphony/engine/tests/test_engine_metadata.py
```diff
@@ -0,0 +1,134 @@
...
+                                             EngineInterface.FileIO])
+
+            return [example_engine]
+
+        def create_wrapper(self, cuds, engine_name, engine_interface):
+            if engine_name == 'EXAMPLE2':
+                pass
+            else:
+                raise Exception('Only EXAMPLE2 engine is supported. '
+                                'Unsupported eninge: %s', engine_name)
```
kitchoi: `Exception("blahblah %s" % engine_name)` ?  (and there is another one on L26)

simphony/engine/tests/test_engine_metadata.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class TestEngineMetadata(unittest.TestCase):
+    """Test everything engine metadata."""
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
+
```
kitchoi: There is no need to define setUp and tearDown if they are empty.

simphony/engine/tests/test_engine_metadata.py
```diff
@@ -0,0 +1,134 @@
...
+    def test_get_supported_engines(self):
+        supported = engine_api.get_supported_engines()
+        assert(isinstance(supported, list))
+
+    def test_create_wrapper(self):
+        pass
```
kitchoi: If this is to be completed, can you please add a FIXME/TODO comment here?  It would be easier to track these TODO's down later.

kitchoi:
I made a few minor comments in the tests.  There is also one single, trivial flake8 error that flags the CI checks red.  After that, good to merge.

kitchoi:
Thanks!

mehdisadeghi:
Thanks @kitchoi ! I'm adding more tests, I will correct the above issues.

simphony/engine/tests/test_engine_metadata.py
mehdisadeghi: > Exception("blahblah %s" % engine_name) ? (and there is another one on L26)

I assume you mean why I have hard-coded these dummy names. Those engines are there only for sake of testing. I do not rely on the other engines (like lammps) to be loaded, that should be part of each specific wrapper's tests.

kitchoi: No, it was just the "%" versus "," :)  really minor

doc/source/extending_simphony.rst
```diff
@@ -56,6 +56,70 @@ Will allow the user to import the new engine from inside the ``simphony`` module
...
+SimPhoNy will inspect the plugins for subclasses of ``ABCEngineExtension``. SimPhoNy
+creates an instance of each class and keeps track of available wrappers and their capabilities.
+As implemented in the above example, there are two methods that should be implemented by
+wrapper developers. ``create_wrapper`` is a factory method that receives a ``CUDS``, an
+engine name and an interface (according to ``EngineInterface`` enum). This method must
+create a wrapper instance and return it.
+
+The other method is ``get_supported_engines``. This method must create and return a list of
+the engine's features. Features are simply a pair of physics equation and a list of methods
+that the corresponing egnine provides in order to solve the physics equation. Features are
```
tuopuu: egnine -> engine

doc/source/extending_simphony.rst
```diff
@@ -56,6 +56,70 @@ Will allow the user to import the new engine from inside the ``simphony`` module
...
+create by calling the ``create_engine_metadata_feature``. After creating features one must
```
tuopuu: create -> created

doc/source/extending_simphony.rst
```diff
@@ -56,6 +56,70 @@ Will allow the user to import the new engine from inside the ``simphony`` module
...
+pass the to the ``create_engine_metadata`` method in order to create metadata objects. Each
```
tuopuu: the -> them

simphony/cuds/model.py
```diff
@@ -0,0 +1,170 @@
...
+        component: CUDSComponent
+            a cuds component
+        """
+        if not isinstance(component_id, (str, uuid.UUID)):
+            raise TypeError('ID should be of string or UUID type')
+        if isinstance(component_id, uuid.UUID):
+            component_id = str(component_id)
+        if component_id in self._map:
+            return self._map[component_id]()
+
```
kitchoi: there should be an else clause here?  (i.e. `component_id in self._map` evaluates to False)

kitchoi: Ah nevermind, suppose you get a KeyError here.

simphony/cuds/model.py
```diff
@@ -0,0 +1,170 @@
...
+        # Add datasets to memory datastore by default
+        self._dataset_store = MemoryStateDataStore()
+
+        # Memory storage to keep CUDS components
+        self._store = {}
+
+        # The generic data container
+        self._data = DataContainer()
+
+        # A map to find the object for a given id amoung datasets or
```
tuopuu: amoung -> among

simphony/cuds/model.py
```diff
@@ -0,0 +1,170 @@
...
+        # containers do not have `uid' and therefore this workaround
+        # is necessary. This is a dict of key of any kind and a lambda
+        # that will return the corresponding value.
+        self._map = {}
+
+    @staticmethod
+    def _is_cuds_component(obj):
+        """Check whether the object is a CUDS component."""
+        # When I see a bird that walks like a duck and swims like a duck
+        # and quacks like a duck, I call that bird a duck
```
tuopuu: :+1: :smile:

simphony/cuds/store.py
```diff
@@ -0,0 +1,91 @@
...
+class MemoryStateDataStore(ABCStateDataStore):
+    """A dataset store that keeps everything in memory."""
+    def __init__(self, *args, **kwargs):
+        self._datasets = {}
+
+    def add(self, dataset, *args, **kwargs):
+        """Add the dataset to the store."""
+        if dataset.name in self._datasets:
+            raise Exception('Dateset %s already exists.' % dataset.name)
```
tuopuu: Dateset -> Dataset

simphony/engine/__init__.py
```diff
@@ -3,9 +3,62 @@
...
+_ENGINE_MANAGER = EngineManager()
+
+
+def get_supported_engine_names():
+    """Show a list of supported engine names.
+
+    Returns
+    -------
+    names: list
+        a list of engine names.
```
tuopuu: remove full stop

simphony/engine/__init__.py
```diff
@@ -3,9 +3,62 @@
...
+    return _ENGINE_MANAGER.get_supported_engines()
+
+
+def create_wrapper(cuds, engine_name, engine_interface=None):
+    """Create a wrapper to the given engine.
+
+    Parameters
+    ----------
+    cuds: CUDS
+        A cuds object which contains model information.
```
tuopuu: I'd suggest remove capital first letter and full stop at the end from these lines. They are not really proper sentences that need those, and it would be consistent with earlier docstrings.

simphony/cuds/model.py
```diff
@@ -0,0 +1,170 @@
...
+            self._map[component_id] = \
+                lambda key=component_id: self._store.get(key)
+
+    def get(self, component_id):
+        """Gets a component from the CUDS computational model.
+
+        Parameters
+        ----------
+        component_id: uuid.UUID
+            key of a CUDS component.
```
tuopuu: remove full stop

simphony/engine/extension.py
```diff
@@ -0,0 +1,219 @@
...
+
+class EngineMetadata(object):
+    """Data structure to represent engine metadata.
+
+    This class represents one supported engine along with its features.
+
+    Parameters
+    ----------
+    name: str
+      fixed name of a support engine to be used in user's data, i.e. key.
```
tuopuu: remove full stop at the end

simphony/engine/extension.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+
+    @staticmethod
+    def create_engine_metadata(name, features, interfaces):
+        """Factory method to create engine metadata.
+
+        Parameters
+        ----------
+        name: str
+          fixed name of a support engine to be used in user's data, i.e. key.
```
tuopuu: remove full stop at the end

simphony/cuds/model.py
mehdisadeghi: :+1:

simphony/engine/extension.py
```diff
@@ -0,0 +1,219 @@
...
+                                     extension)
+                self.add_extension(extension)
+
+    def add_extension(self, extension):
+        """Add an extension to the context.
+
+        Parameters
+        ----------
+        extension: engine.ABCEngineExtension
+          an extension that has knowledge about its own engines.
```
tuopuu: remove full stop

doc/source/extending_simphony.rst
mehdisadeghi: Thanks, I was writing them correctly in my mind, my finger's mistake!

simphony/engine/extension.py
```diff
@@ -0,0 +1,219 @@
...
+            # It is our interaction point with its engines.
+            self._engine_extensions[engine.name] = extension
+
+    def create_wrapper(self, cuds, engine_name, engine_interface=None):
+        """Create a wrapper to the given engine.
+
+        Parameters
+        ----------
+        cuds: CUDS
+            A cuds object which contains model information.
```
tuopuu: Again probably better to change these description lines to not have a capital first letter and a full stop at the end.

simphony/cuds/model.py
mehdisadeghi: I call that object a CUDS component! :D

tuopuu:
Good work @mehdisadeghi . I just only found a few typos in the docstrings. Otherwise, good to merge! :smile:

simphony/cuds/model.py
kitchoi: so this method returns None if component_id refers to a component not previously added.

mehdisadeghi:
@kitchoi @tuopuu Thanks for the reviews! I'll merge after successful build.

simphony/engine/__init__.py
mehdisadeghi: Yes, removed many of them!

simphony/cuds/model.py
kitchoi: After reading the code a second time, am I right that if you add a dataset (ABCParticles, ABCLattice, ABCMesh) to the CUDS, you can't get it back from the CUDS using its name (since they don't have uid)?

mehdisadeghi: Yes, true. I try to stay close the python implementation of `dict`.

The day I get to know the "Thao of Programming" in full depth, ultimately, I will know by heart,  which signature is the best, without looking at the code, in advance! :smile:

kitchoi: Now this works.  But are we going to keep it this way in the future or are we going to have uid for datasets too?
```
lattice = Lattice(name="lattice_name"...)
CUDS.add(lattice)
CUDS.get("lattice_name")
```

simphony/cuds/model.py
```diff
@@ -0,0 +1,170 @@
...
+                    hasattr(obj, 'data')])
+
+    @staticmethod
+    def _is_dataset(obj):
+        """Check if the object is a dataset."""
+        if isinstance(obj, (ABCParticles, ABCLattice, ABCMesh)):
+            return True
+        else:
+            return False
+
```
kitchoi: :+1:
even better:
`return isinstance(obj, (ABCParticles, ABCLattice, ABCMesh))`

simphony/cuds/model.py
mehdisadeghi: Good question. Since datasets do not have uid, I had to implement a workaround for now (not to duplicate everything just for them). 

So, probably yes, we will add uid to datasets too.

simphony/cuds/model.py
mehdisadeghi: Of course!

simphony/cuds/model.py
kitchoi: Thanks!  Good to know :+1:

PR #288, opened 2016-04-25
Feature: Create new uuid when add component to CUDS

kitchoi:
Consistent with Particles and Mesh, new uuid can be created when a component to be added does not already have one (e.g. `component.uid is None`).

Alternatively, we can make sure every CUDS component has a uuid created upon `__init__`, but it does not seem to me necessary.

Note that I have removed converting `uuid` to `str` as `uuid` is already hashable (a pretty good hash) so we don't need to convert it to str to be hashed again.

The type checking for the component_id in the `add`, `get` and `remove` functions wasn't symmetric (i.e. it is possible to add something you can't `remove` later, see FIXME comment).  Here I removed these type checking altogether, but we might want to add them back.

### DIFF
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -65,14 +65,42 @@ class CUDS(object):
     def add(self, component):
         """Add a component to the CUDS computational model.
 
+        If the component has `uid` attribute but that value is
+        None, a new uuid will be created and assigned to the
+        component after the component is successfully added
+        to the CUDS computational model.
+
         Parameters
         ----------
         component: CUDSComponent
             a component of CUDS, reflecting SimPhoNy metadata
+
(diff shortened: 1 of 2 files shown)


mehdisadeghi:
@kitchoi You have removed the answer to the life universe and everything from the tests!

kitchoi:
@mehdisadeghi ahaha yes, because cuds.get(42) gives you `None`, and I consider this tested already.

mehdisadeghi:
@kitchoi does this PR needs a corresponding issue?

kitchoi:
@mehdisadeghi  Could do.

mehdisadeghi:
@kitchoi nevermind. Just merged.

PR #292, opened 2016-05-04
Problem: implicit code inspection for metadata loading is unclear and complex

mehdisadeghi:
Solution: make it explicit by adding a `register` decorator.


mehdisadeghi:
@kitchoi @tuopuu Today I ran out into another problem while loading extensions so I came up with a new solution. I added a new class decorator called 'register' and merged `extension` and `engine` packages (less clutter). Please have a look, thanks.

kitchoi:
@mehdisadeghi  Are there now two ways to add an engine extension to an EngineManger?  One by calling `EngineManger.load_metadata(module)`, another by the `register` decorator which add the extension when the module is being loaded?
What should happen if I have already loaded the module that contains an engine extension (i.e. registered to an EngineManager by the decorator) but then I want to replace the EngineManager instance with another EngineManager? (I think that's the intention, right?)

kitchoi:
@mehdisadeghi  I can't find where `set_engine_manager` is used.  What's its use case?  How do we retrieve the extensions registered to the EngineManager that is replaced *by* the `set_engine_manager`?

kitchoi:
My comment regarding replacing the `EngineManager` after importing an extension-containing module can be shown in this test case here: https://github.com/simphony/simphony-common/blob/fix-cyclic-imports-experiment/simphony/engine/tests/test_register_engine_extension.py

mehdisadeghi:
@kitchoi I wrote those tests and `set_engine_manager` having implicit load in mind, hence some inconsistency.

I'll update the code to better reflect the use of the decorator.

mehdisadeghi:
@kitchoi Please checkout the latest changes. Feel free to add your changes in a commit or PR.

kitchoi:
Thanks @mehdisadeghi.  The changes look good.  There is an error in the tests, can you fix that please?

mehdisadeghi:
I have already fixed that, will push it soon.

mehdisadeghi:
@tuopuu @SGGgarcia @kitchoi @roigcarlo Please review and press the merge button if is OK!

kitchoi:
Looks good to me

Issue #223, opened 2015-11-10
`supported_cubas` optional initialization argument for Mesh/Lattice/Particles


nathanfranklin:
This issue covers the optional initialization argument for Mesh/Lattice/Particles from @itziakos :  https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/Supported_CUBA_keys_for_container .  (Note that the other points from this proposal are covered in #217)
- The Mesh, Particles and Lattice objects should accept at initialization an optional argument passing a set of cuba values that will be supported. Default values for the supported cuba keys for each container type should be decided by the related workgroups and the SSB. 

```
mesh1 = Mesh(name='test1', supported_cubas={CUDSItem.POINT:set(CUBA.VELOCITY), CUDSItem.CELL:set(CUBA.COLOR)})
```

Description of what needs to be done is being disussed also on the wiki.

stefanoborini:
No longer relevant. We don't constrain anymore.

Issue #217, opened 2015-11-04
supported_cuba and contained_cuba methods for CUDS datasets 


nathanfranklin:
from @itziakos :  https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/Supported_CUBA_keys_for_container

kitchoi:
I think this is related:
In [abc_check_mesh](https://github.com/simphony/simphony-common/blob/36e8a752250b0495f5ee03ff6f42d48368b1bcf9/simphony/testing/abc_check_mesh.py#L414), should `self.supported_cuba()` be supplied too? (i.e. `create_points(restrict=self.supported_cuba()`)

nathanfranklin:
@kitchoi ,  I totally agree.  the testing frameworks (`abc_check_mesh`, `abc_check_lattice`..) should be updated to make use of the new `supported_cuba(item_type)` methods

ahashibon:
Since the SSB is finalising the CUDS design, this should be put on hold until we define the CUDS more accurately, please remove from the current milestone.

Issue #216, opened 2015-10-29
ABCCheckLattice assumes that e.g. get_node returns empty LatticeNodes


tuopuu:
ABCheckLattice template assumes that a lattice initiated for testing has no CUBA data stored in it's LatticeNode objects. If the lattice is initiated in the container_factory method with, say, CUBA.MATERIAL_ID defined for all nodes, then CheckLatticeNodeOperations.test_get_node causes an error because it assumes that the returned LatticeNode (node) has an empty DataContainer object as node.data.

JYULatticeProxy lattice has by default certain CUBA keywords defined for the lattice, so using a get_node always returns LatticeNodes that have non-empty DataContainer object accompanied with it. This is currently restricting us of using the test template for JYULatticeProxy.

I propose that ABCheckLattice template is modified so that it does not expect empty nodes returned from get_node, and methods alike. The change can be done by modifying the template, or compare_data_containers and compare_lattice_nodes methods from simphony.testing.utils.py

nathanfranklin:
> I propose that ABCheckLattice template is modified so that it does not expect empty nodes returned from get_node, and methods alike. The change can be done by modifying the template, or compare_data_containers and compare_lattice_nodes methods from simphony.testing.utils.py
 
Could the fix just be changing the following line https://github.com/simphony/simphony-common/blob/master/simphony/testing/abc_check_lattice.py#L183 from:

`node.data = create_data_container() `
to
` node.data = create_data_container(restrict=self.supported_cuba())`

> This is currently restricting us of using the test template for JYULatticeProxy.

As a work-around, could you still use `CheckLatticeNodesOperations` and just override the `test_get_node` method (like https://github.com/simphony/simphony-lammps-md/blob/14474e5e118a9233d9c4331b445c2ccd50bec6f0/simlammps/tests/test_particles.py#L66_L80) so that you can still use the current version ABCCheckLattice testing template

tuopuu:
@nathanfranklin, thanks for the ideas!

> Could the fix just be changing the following line

Unfortunately this breaks the template for other lattice classes, Lattice and H5Lattice.

I have tried to avoid overriding for now, because I would have to override quite many tests if I went that road. But I'm still testing other ways of solving this. One of the other ways might be to get rid of JYULatticeProxy for good and replace it with a regular Lattice class, that is stored internally in the engine.

[...]

PR #293, opened 2016-05-16
Feature: Add support for restricted cuba key in DataContainer

kitchoi:
In order to replicate the keyword checking in `MaterialRelation` (https://github.com/simphony/simphony-common/blob/master/simphony/cuds/material_relations/material_relation.py#L107) for all the generated classes from simphony-metadata, it would be more effective if the checking is done at the `DataContainer `level.

For example:
```
>>> from simphony.core.data_container import create_data_container
>>> container = create_data_container((CUBA.NAME, CUBA.VELOCITY))()
>>> container.restricted_keys
frozenset({<CUBA.VELOCITY: 55>, <CUBA.NAME: 61>})
>>> container[CUBA.NAME] = 'name'
>>> container[CUBA.POSITION] = (1.0, 1.0, 1.0)
...
ValueError: Key <CUBA.POSITION: 119> is not in the supported CUBA keywords
```

Performance of the DataContainer is basically the same to the one we have now.  Benchmark:
```
Initialization:
dict: 1000000 calls, best of 5 repeats: 0.000001 sec per call
DataContainer: 100000 calls, best of 5 repeats: 0.000011 sec per call  (was 0.000010)
dict == DataContainer True

Iterations:
dict: 1000000 calls, best of 5 repeats: 0.000001 sec per call
DataContainer: 1000000 calls, best of 5 repeats: 0.000001 sec per call  (same)

getitem access:
dict: 100000 calls, best of 5 repeats: 0.000007 sec per call
DataContainer: 100000 calls, best of 5 repeats: 0.000012 sec per call  (same)
dict == DataContainer True

setitem with CUBA keys:
dict: 10000 calls, best of 5 repeats: 0.000024 sec per call
DataContainer: 10000 calls, best of 5 repeats: 0.000061 sec per call  (was 0.000058)
dict == DataContainer True
```
(Running with Ubuntu 12 on a VM with 1 core, 2.5GHz, 4GB RAM)

Please consider this as a proposal and please comment, review or propose alternatives.

### DIFF
--- a/simphony/core/data_container.py
+++ b/simphony/core/data_container.py
@@ -1,3 +1,5 @@
+import sys as _sys
+
 from simphony.core.cuba import CUBA
 
 
@@ -105,6 +107,12 @@ def create_data_container(restricted_keys):
     ''' Create a DataContainer subclass with the given
     restricted keys
 
+    Note
+    ----
+    For pickling to work, the created class needs to be assigned
+    a name `RestrictedDataContainer` in the module where it is
+    created
+
     Parameters
     ----------
     restricted_keys : sequence
@@ -119,7 +127,7 @@ def create_data_container(restricted_keys):
     --------
     >>> container = create_data_container((CUBA.NAME, CUBA.VELOCITY))()
     >>> container.restricted_keys
-    {<CUBA.VELOCITY: 55>, <CUBA.NAME: 61>}
+    frozenset({<CUBA.VELOCITY: 55>, <CUBA.NAME: 61>})
     >>> container[CUBA.NAME] = 'name'
     >>> container[CUBA.POSITION] = (1.0, 1.0, 1.0)
     ...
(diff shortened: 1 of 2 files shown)


kemattil:
Restricting keywords for a data container (e.g. by specifying a fixed set of supported keywords at the initialization) is something we have been discussing for a long time now and is related to several aspects of the SimPhoNy design (see at least issues #29, #216, #217, #223).

I continue to favour implementing this feature.

In addition, the proposed implementation with a restricted subclass of the data container and utilization of the frozenset seems elegant.

mehdisadeghi:
It seems to me that `DataContainer` and `CUDSItem` are very similar and serve similar purposes. Why not consider DataContainer as the base class for metadata classes?

mehdisadeghi:
I'm in favor of type and consistency checking in metadata classes and CUDS itself. IMO the functionality of `create_data_container` should be merged in generated classes.

kitchoi:
> IMO the functionality of create_data_container should be merged in generated classes.

Do you mean using `create_data_container` in the generated class or moving the code (`def create_data_...`) to the generated class module?
My intention is for the former, so that with, say, the MaterialRelation class:
```
>>> material_relation = MaterialRelation()

>>> # same as material_relation.supported_parameters, but a set
>>> material_relation.data.restricted_keys
frozenset({CUBA.MATERIAL})

>>> material_relation.data[CUBA.POSITION] = (1.0, 1.0, 1.0)
ValueError: Key <CUBA.POSITION: 119> is not in the supported CUBA keywords
```

mehdisadeghi:
Your latest code example is close to what I intended and looks good to me. As long as the use of the factory method is internal, it is fine.

kitchoi:
Just added an optional argument to the `create_data_container` function for setting class name so that one can assign the dynamically created class to a (conventionally) private variable while still have pickling support.

PR #294, opened 2016-05-25
Problem: CUDS.iter is broken.

mehdisadeghi:
Solution: Consider DataSet classes and fix it.


kitchoi:
How was it broken?  Could you add a test that capture the broken scenario please?

mehdisadeghi:
The test was there, but it was not capturing this scenario: Having added some datasets, iter was returning an empty generator.

PR #297, opened 2016-06-09
Reduced equation of state -tools

kemattil:
**Motivation/Problem**
Two phase (liquid-vapour) fluid flow simulations need to be initialized with good/sensible values for the density/pressure fields. These values depend on the equation of state chosen for the modeled fluid (a materials relation) as well as on the system temperature (a system parameter).

**Note**
This is a general modelling problem (theory of thermodynamics), not a particular difficulty of, e.g., JYULB-solver.

**Solution**
Implementation of a (general) tool which computes liquid-vapour densities based on the given equation of state and system temperature.

**Features**
- Several equations of state implemented in reduced form (`reos.py`). Reduced equations of state are general in the sense that they do not involve material parameters (e.g. physico-chemical constants)
which characterize a particular, real fluid.
- Computation of liquid-vapour densities by utilizing the Maxwell's equal area rule (`maxwell_equal_area_rule.py`).
- Unit tests (`test_eos_tools.py`).

**Dependencies**
- requires scipy (`dev_requirements.txt` updated)

@tuopuu @kitchoi @mehdisadeghi @roigcarlo please have a look, if possible.


simphony/tools/maxwell_equal_area_rule.py
```diff
@@ -0,0 +1,240 @@
...
+
+def liquid_vapour_density(Trs, reos, iSearchMax=40.0, RhorMin=1e-12):
+    """Compute liquid-vapour densities for the given reduced temperatures.
+
+    The densities are computed by utilizing the Maxwell's equal area rule.
+
+    Parameters
+    ----------
+    Trs : Reduced temperatures (iterable)
+    reos : Reduced equation of state
```
tuopuu: This can be just
"""
   reos : ReducedEquationOfState
"""

simphony/tools/maxwell_equal_area_rule.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def liquid_vapour_density(Trs, reos, iSearchMax=40.0, RhorMin=1e-12):
+    """Compute liquid-vapour densities for the given reduced temperatures.
+
+    The densities are computed by utilizing the Maxwell's equal area rule.
+
+    Parameters
+    ----------
+    Trs : Reduced temperatures (iterable)
```
tuopuu: I think it's better to define parameter types explicitly (iterable of int, float, string etc.) and add a description of its use on the next line. In this case it could be
"""
   Trs : iterable of float
           Reduced temperatures
"""

tuopuu:
@kemattil, We should find out what's the problem with travis-ci in this branch, and fix it. Other than that, I would like to change most of the parameter descriptions in the methods in the format we've been using in SimPhoNy: 
```
parameter_name : type
                              Descriptive textual definition
```
I understand that in most cases in this branch, the type would be just a float, but it's better to keep documentation style consistent throughout the project.

About SciPy dependency: I think it's unlikely that simphony-common itself would need SciPy, so we might want to separate it in dev_requirements.txt file. It's likely that when we start adding more physics related tools to simphony/tools we have to add other dependencies as well (such as linear algebra libraries etc). I suggest that all dependencies of simphony/tools would be grouped together either in a separate file, or below a comment line (something like # Requirement for simphony/tools ) in dev_requirements.txt.

tuopuu:
@kemattil, we had a discussion in the DevTeam meeting about adding tools to SimPhoNy and decided to build a separate repository (simphony-tools) for those. This way we avoid cluttering simphony-common requirements list with packages that are very specific to certain tools and are not needed elsewhere inside SimPhoNy.

Do you mind if I take control of this branch and add it there?

kemattil:
@tuopuu thanks for the comments!

A separate simphony-tools repository is a good idea.

Take control of the branch by all means!

kemattil:
@tuopuu I made the changes to parameter descriptions as you suggested.

Still do not know why the the Travis CI fails ...

tuopuu:
Thanks @kemattil. Code and descriptions now look good. It seems that AppVeyor error is caused by SciPy not installed in AppVeyor script, which we can fix when moving these files to simphony-tools repository. Travis-CI fail seems unrelated to this branch.

kemattil:
Ok, good.

kitchoi:
@tuopuu @kemattil The Travis CI failure is due to failing to automatically generate the documentation using the docstrings provided.  There are a number of long docstrings in this branch which might cause problems for sphinx.  Please try `python setup.py build_sphinx` to test locally if the doc can be generated successfully.

tuopuu:
This branch has been moved to simphony-tools repository.

PR #300, opened 2016-06-24
Import meta generated cuds and keywords

mehdisadeghi:
@roigcarlo @kitchoi @SGGgarcia @tuopuu 

Please review this initial PR and go through the failed tests and see how we can fix them.


setup.cfg
```diff
@@ -1,2 +1,3 @@
 [flake8]
-exclude = doc/*
+# Exclude docs and generated files.
+exclude = doc/*,simphony/cuds/meta/,simphony/core/keywords.py,simphony/core/cuba.py
```
itziakos: There is not real reason why the generated python code is not pep8 compatible. This should be fixed.

kitchoi:
@mehdisadeghi @itziakos Good to merge?

kitchoi:
The generated code decreases coverage by a lot!

kitchoi:
Coverage could be enhanced if we bring in the tests in `simphony-metadata`

mehdisadeghi:
Thanks for all the fixes @kitchoi ! I also saw your other branch, I'll regenerate.

mehdisadeghi:
@kitchoi @itziakos Any idea why does the appveyor test fail?

kitchoi:
@mehdisadeghi I believe it has to do with a new version flake8, here is the relevant issue reported on flake8: https://gitlab.com/pycqa/flake8/issues/179

mehdisadeghi:
@kitchoi This is a symptom of the problem that I had described [before](https://github.com/simphony/simphony-framework/issues/15).

kitchoi:
@mehdisadeghi In my last commit, I skipped the known problematic versions (3.0.0, 3.0.1) of flake8.  If new flake8 version is made available, e.g. 3.0.2, pip would still install the latest version.  It is possible that the known issue could be fixed on 3.0.2, and it would be better that we install it.  If 3.0.2 comes up and the bug is not fixed, we skip it again.

mehdisadeghi:
@kitchoi Even though the version of pip on appveyor is the same as the one on my local machine (8.1.2), surprisingly, it does not recognize the exclusion operator that you've added:

```
50 Invalid requirement: 'flake8=3.0.1'
51 = is not a valid operator. Did you mean == ?
```

mehdisadeghi:
I think we can proceed with merging this PR.

PR #310, opened 2016-07-28
Problem: imports in simphony.__init__ may cause import cycles

### LINKED ISSUE #301: Imports in simphony.__init__ may cause import cycles
If an engine wants to import anything under simphony, the __init__ would import, e.g. Simulation, which imports `engine`.  However `engine` has not been imported and this causes an import cycle.

A simple fix and the recommended approach is to keep simphony/__init__.py empty.

mehdisadeghi:
Solution: move public API (imports in simphony.__init__.py) to simphony.api

Close #301


mehdisadeghi:
@SGGgarcia @kitchoi please review the changes.

kitchoi:
Looks good to me.

PR #311, opened 2016-08-01
Add new entry point for pre processing

SGGgarcia:
This PR adds a new entry point for stevedore called `simphony.pre_processing` currently used by nFluid tool. Also it will be potentially used for any other pre_processing tool.


itziakos:
Please update the documentation as necessary https://github.com/simphony/simphony-common/blob/master/doc/source/extending_simphony.rst

PR #312, opened 2016-08-02
Update cuds api according to m30

mehdisadeghi:
This PR simplifies CUDS class. Adds `name` and `description` to `CUDS` and `Simulation` classes. Moreover, uses `name` as the main parameter to add/get/remove items inside CUDS.

### DIFF
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -147,7 +147,15 @@ class CUDS(object):
         -------
         component: CUDSComponent
             the corresponding CUDS component or None
+
+        Raises
+        ------
+        TypeError
+            if the name is not a non empty string
         """
+        if not name:
+            raise TypeError('name must be a non empty string.')
+
         uid = self._name_to_id_map.get(name)
         return self.get_by_uid(uid)
 
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -115,16 +115,9 @@ class CUDS(object):
             raise TypeError('Not a CUDSComponent nor a dataset object: %s.' %
                             type(component))
 
-        # Do not accept items with duplicate names unless the expected
-        # behaviour is defined clearly for the following sample code:
-        # m1 = api.Material(name="item1")
-        # m2 = api.Material(name="item1")
-        # b1 = api.Box(name="item1")
-        # b2 = api.Box(name="item1")
-        # cuds.add(m1)
-        # cuds.add(m2)
-        # cuds.add(b1)
-        # cuds.add(b2)
+        # Do not accept items with duplicate names.
+        # Components/datasets with no name will be added, however
+        # it is not possible to get/remove them with `name`
         if component.name in self._name_to_id_map:
             raise ValueError('Name clash. A %s component has the same name %s'
                              % (type(self.get(component.name)),
@@ -140,8 +133,7 @@ class CUDS(object):
             self._map[component.name] = \
                 lambda key=component.name: self._dataset_store.get(key)
             # Datasets use name as id
-            if component.name not in (None, ''):
-                self._name_to_id_map[component.name] = component.name
+            self._name_to_id_map[component.name] = component.name
         else:
             # Delete existing item with the same name
             # Store the component. Any CUDS item has uid property.
@@ -149,6 +141,9 @@ class CUDS(object):
             self._map[component.uid] = \
                 lambda key=component.uid: self._store.get(key)
             # Store name for name-to-id mapping.
+            # Components with no name will are not added to the mapping
+            # and it is not possible to access them using add/remove
+            # methods. Use `get_by_uid` and `remove_by_uid` instead.
             if component.name not in (None, ''):
                 self._name_to_id_map[component.name] = component.uid
 
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -62,14 +62,6 @@ class CUDS(object):
     def uid(self):
         return self._uid
 
-    @uid.setter
-    def uid(self, value):
-        raise TypeError("uid is readonly.")
-
-    @uid.deleter
-    def uid(self):
-        raise TypeError("Can't delete uid.")
-
     @staticmethod
     def _is_dataset(obj):
         """Check if the object is a dataset."""
@@ -86,10 +78,6 @@ class CUDS(object):
     @data.setter
     def data(self, value):
         self._data = DataContainer(value)
-
-    @data.deleter
-    def data(self):
-        raise TypeError("Can't delete data.")
 
     def add(self, component):
         """Add a component to the CUDS computational model.
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -132,6 +132,9 @@ class CUDS(object):
 
         # Add datasets separately
         if self._is_dataset(component):
+            if not component.name:
+                raise TypeError('Dataset must have a name.')
+
             self._dataset_store.add(component)
             # Datasets at the moment do not have uid
             self._map[component.name] = \
@@ -193,7 +196,13 @@ class CUDS(object):
         ------
         KeyError
             if no component exists of the given name
+
+        TypeError
+            if the name is not a non empty string
         """
+        if not name:
+            raise TypeError('name must be a non empty string.')
+
         uid = self._name_to_id_map.get(name)
         self.remove_by_uid(uid)
 
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -3,6 +3,8 @@
 This module contains the classes used to represent a computational model,
 based on SimPhoNy metadata.
 """
+import uuid
+
 from .store import MemoryStateDataStore
 from .abc_particles import ABCParticles
 from .abc_lattice import ABCLattice
@@ -28,6 +30,9 @@ class CUDS(object):
     """
     def __init__(self, name=None, description=None):
 
+        # Assign unique ID
+        self._uid = uuid.uuid4()
+
         # Add name
         self.name = name
 
@@ -52,6 +57,18 @@ class CUDS(object):
 
         # Another map to keep a mapping between names and ids
         self._name_to_id_map = {}
+
+    @property
+    def uid(self):
+        return self._uid
+
+    @uid.setter
+    def uid(self, value):
+        raise TypeError("uid is readonly.")
+
+    @uid.deleter
+    def uid(self):
+        raise TypeError("Can't delete uid.")
 
     @staticmethod
     def _is_dataset(obj):
--- a/simphony/cuds/model.py
+++ b/simphony/cuds/model.py
@@ -120,7 +120,8 @@ class CUDS(object):
             self._map[component.name] = \
                 lambda key=component.name: self._dataset_store.get(key)
             # Datasets use name as id
-            self._name_to_id_map[component.name] = component.name
+            if component.name not in (None, ''):
+                self._name_to_id_map[component.name] = component.name
         else:
             # Delete existing item with the same name
             # Store the component. Any CUDS item has uid property.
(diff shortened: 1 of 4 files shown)


mehdisadeghi:
@tuopuu @kitchoi @SGGgarcia Please review this PR.
@ahashibon Please see the changes regarding `name` parameter.

simphony/cuds/simulation.py
```diff
@@ -14,8 +14,17 @@ class Simulation(object):
...
+    description: str
+        More information about this simulation
     """
-    def __init__(self, cuds, engine_name, engine_interface=None):
+    def __init__(self, cuds, engine_name, engine_interface=None,
+                 name=None,
+                 description=None):
+        self.name = name
+        self.description = description
+
```
kitchoi: `Simulation` does not have `__slots__`.  User can always do:
```
sim = Simulation()
sim.name = 'sim1'
sim.description = 'description for sim1'
```
Or, are the names/descriptions required in the API (then they should be default to be None)?

mehdisadeghi: They are optional at the moment, exactly like the meta classes.  Perhaps we have to turn `CUDS` and `Simulation` classes into `CUDSItem` derivaties, to have one universal data structure to deal with (it has been always the aim). However, for now I just decided to keep it simple.

kitchoi:
Looks like with this PR makes it necessary to define a unique string `name` for every item being added to a CUDS.  `name` essentially replaces uuid.   Is this right?  If so, is this behaviour intended by the SSB?

simphony/cuds/tests/test_cuds.py
```diff
@@ -11,14 +11,8 @@ class CUDSTestCase(unittest.TestCase):
...
 
-        # TODO: use generated components
-        class DummyComponent(object):
-            def __init__(self):
-                self.uuid = uuid.uuid4()
-                self.name = 'dummyname'
-                self.data = {}
-        self.dummpy_component1 = DummyComponent()
-        self.dummpy_component2 = DummyComponent()
+        self.dummpy_component1 = api.Box(name='mybox')
```
tuopuu: While at it, should fix this to self.dummy_component1 & 2. :)

mehdisadeghi:
> name essentially replaces uuid. Is this right?

To some extent, yes. `uid` for internal consistency, `name` for users.

> is this behaviour intended by the SSB

AFAIK, Yes. Hence mentioning Adham to confirm.

As I have commented in the code, the behavior of add/get/remove by name vs. by uid is ambiguous. What happens if a user adds a `Box` and `Material` with the exact same names? The initial behavior was to replace them blindly (as in python dict), I forced it to reject duplicates to have further discussions.

simphony/cuds/tests/test_cuds.py
mehdisadeghi: Oh gotcha! I'll fix them right away! :)

simphony/cuds/simulation.py
ahashibon: I agree with @mehdisadeghi  here. For now it should be enough.

ahashibon:
>> As I have commented in the code, the behavior of add/get/remove by name vs. by uid is ambiguous. What happens if a user adds a  Box  and  Material  with the exact same names? The initial behavior was to replace them blindly (as in python dict), I forced it to reject duplicates to have further discussions.
For now this is ok. Next iteration we should allow the same name for different items. For example a Box and Material may have the same name, but two different materials cannot. In essence this would require a way in the API to distinguish between the items.

mehdisadeghi:
@tuopuu @kitchoi @SGGgarcia based on our discussions I made some changes. Please have a look.

simphony/cuds/model.py
```diff
@@ -39,119 +55,179 @@ def __init__(self):
...
-                    hasattr(obj, 'data')])
+        # Another map to keep a mapping between names and ids
+        self._name_to_id_map = {}
+
+    @property
+    def uid(self):
+        return self._uid
+
+    @uid.setter
+    def uid(self, value):
```
kitchoi: If you don't provide the setter, it is read-only.  Trying to set it would raise an `AttributeError: can't set attribute`

simphony/cuds/model.py
```diff
@@ -39,119 +55,179 @@ def __init__(self):
...
+        raise TypeError("uid is readonly.")
+
+    @uid.deleter
+    def uid(self):
```
kitchoi: Likewise, if you don't provide the deleter for a property, you can't delete it.

simphony/cuds/model.py
```diff
@@ -39,119 +55,179 @@ def __init__(self):
...
-        # assigns the new uid to the component.uuid
-        # Only do so after successfully adding the component
-        if hasattr(component, "uuid"):
-            component.uuid = component_id
+        Returns
+        -------
+        component: CUDSComponent
+            the corresponding CUDS component or None
+        """
+        uid = self._name_to_id_map.get(name)
```
kitchoi: If name is not found, `uid` would be None.  What happens is that the cuds item cannot be retrieved by name, so it would make sense to raise an Error here directing the user to use `get_by_uid`

kitchoi: > it would make sense to raise an Error here directing the user to use get_by_uid

or `remove_by_uid`

simphony/cuds/model.py
```diff
@@ -39,119 +55,179 @@ def __init__(self):
...
+                self._is_dataset(component)):
+            raise TypeError('Not a CUDSComponent nor a dataset object: %s.' %
+                            type(component))
+
+        # Do not accept items with duplicate names.
+        # Components/datasets with no name will be added, however
+        # it is not possible to get/remove them with `name`
+        if component.name in self._name_to_id_map:
+            raise ValueError('Name clash. A %s component has the same name %s'
+                             % (type(self.get(component.name)),
```
kitchoi: I would skip `self.get(component.name)` here to avoid confusing error being raised from `self.get` etc.

simphony/cuds/model.py
kitchoi: Or something similar to `remove` is sufficient:
```
+        if not name:
+            raise TypeError('name must be a non empty string.')
```

kitchoi:
Is there a `test_get_nameless_component_by_name` version of `test_remove_nameless_component_by_name`?

simphony/cuds/model.py
```diff
@@ -39,119 +55,179 @@ def __init__(self):
...
 
         Raises
         ------
         TypeError
-            if component is not a CUDS component
-
+            if the component is not a CUDS component
         ValueError
-            if the component is already added
+            if a component already exists
```
kitchoi: if a component *with the same name* already exists?

simphony/cuds/model.py
```diff
@@ -39,119 +55,179 @@ def __init__(self):
...
+
     def add(self, component):
         """Add a component to the CUDS computational model.
 
-        If the component has `uid` attribute but that value is
-        None, a new uuid will be created and assigned to the
-        component after the component is successfully added
-        to the CUDS computational model.
+        This will replace any existing value with the same `uid`
+        or `name`.
```
kitchoi: Replace existing value with the same uid, not name

kitchoi:
Please ignore the comment on the existence of test_get_nameless_component_by_name, i saw the test that i was looking for that is named differently.

simphony/cuds/tests/test_cuds.py
```diff
@@ -1,125 +1,223 @@
...
+                         self.nameless_cuds_1)
+        self.assertIsNone(c.get(self.nameless_cuds_1.name))
+
+    def test_get_nameless_cuds_component(self):
+        c = CUDS()
+        c.add(self.nameless_cuds_1)
+
+        self.assertEqual(c.get_by_uid(self.nameless_cuds_1.uid),
+                         self.nameless_cuds_1)
+        self.assertIsNone(c.get(self.nameless_cuds_1.name))
```
kitchoi: I guess my comment in CUDS.get is about whether it is a better user-experience to raise an Error here if name is None or empty.

kitchoi:
Thanks @mehdisadeghi!  Tests are diligent.  There are a few minor issues above, except for that, the rest looks good to merge to me.
Side note: I remain puzzled by the existence of CUDS.data, but I think that's a separate issue/question.

simphony/cuds/model.py
itziakos: I agree, please remove the `setter` and `deleter`  readonly properties only need the `property` decorated function only

simphony/cuds/model.py
```diff
@@ -39,119 +55,179 @@ def __init__(self):
...
-        return isinstance(obj, (ABCParticles, ABCLattice, ABCMesh))
+        return isinstance(obj, (ABCParticles,
+                                ABCLattice,
+                                ABCMesh,
+                                api.DataSet))
 
     @property
     def data(self):
         """Data container for CUDS items."""
         return DataContainer(self._data)
```
itziakos: What is the reasoning for creating a (shallow) copy of the `self._data`. If the purpose is to have a snapshot then this will work only if the contained `keys` and `values` are immutable. In any other case we have an object that references internal information that can change without anybody knowing about them.

If I am correct only the engine functions and returned dataset are required support the snapshot principle. Does this extent to simulation? what is there a reason for this?

simphony/cuds/model.py
```diff
@@ -39,119 +55,179 @@ def __init__(self):
...
+        raise TypeError("Can't delete data.")
+
     def add(self, component):
         """Add a component to the CUDS computational model.
 
-        If the component has `uid` attribute but that value is
-        None, a new uuid will be created and assigned to the
-        component after the component is successfully added
-        to the CUDS computational model.
+        This will replace any existing value with the same `uid`
```
SGGgarcia: Should we add an `update` method so we keep consistent with the rest of simphony and we keep functionality separated?

simphony/cuds/model.py
mehdisadeghi: Good to know that. I'll remove setters and deleters.

simphony/cuds/model.py
mehdisadeghi: This was indeed on purpose. Trying to `get` a nonexistent object yields nothing, like python dict. User can always check the returned value, I think it is more elegant compared to catching an exception. However, since `remove` is not supposed to return anything, an error is desirable in case of a nonexistent key.

simphony/cuds/model.py
mehdisadeghi: I'll remove this one as well.

simphony/cuds/model.py
mehdisadeghi: Corrected.

simphony/cuds/model.py
mehdisadeghi: Yes, corrected.

simphony/cuds/model.py
kitchoi: I don't think `CUDS` can be compared directly with `dict`.  If `dict.get` returns None, we know that the key-value pair does not exist in the `dict`.  However when `CUDS.get` returns `None`, it does not imply there does not exist an CUDS item with an empty name.

simphony/cuds/model.py
mehdisadeghi: Relaced it with uuid.

simphony/cuds/model.py
mehdisadeghi: You are correct, this will not work properly with our current mutable CUDS objects. I think the behavior of the `data` property regarding meta classes is unclear.

I think `Simulation` can have the same API as `CUDS`.

simphony/cuds/model.py
mehdisadeghi: What would be the function of `update` method regarding meta classes?

simphony/cuds/model.py
mehdisadeghi: Right, we are inconsistent here. I added the same check as in `remove` to fix that.

simphony/cuds/model.py
mehdisadeghi: @SGGgarcia Let's discuss this more in detail next time. I suggest to follow it in a separate issue.

Issue #280, opened 2016-02-29
Move Yml files and generatorsn to simphony-metadata


roigcarlo:
Yml files and generators should be moved from simphony-common simphony-metadata.

This links with https://github.com/simphony/simphony-metadata/issues/1

roigcarlo:
The procedure will be:
- [ ] Wait until all yml files are working on simphony-metadata
- [ ] Delete the ones that exist here.

PR #317, opened 2016-08-11
Remove legacy metadata and its generated code.

### LINKED ISSUE #295: Material_relation needs to be replaced
Material Relation class needs to be deleted from simphony-common:
* Mark it as deprecated in the next release
* Delete it after that

mehdisadeghi:
The former material relation code along with its yaml files and its generator is not necessary anymore. We move them partially to the metadata repository. Fix #295 #280

Regarding the new changes to metadata, these classes are redundant and should be removed. I move the yaml files and the generator the the metadata repository.

@tuopuu @kitchoi @SGGgarcia Please see the changes.


stefanoborini:
:+1: when green

PR #318, opened 2016-08-19
Fixes incorrect call in _add_mesh

stefanoborini:
I suspect this is supposed to be add_edges...


roigcarlo:
Indeed, copy paste error :(

stefanoborini:
@roigcarlo which means that some tests are missing in there. I'll fix that, please don't merge.

roigcarlo:
Sure, I am taking a look at it as well. 

If I am not mistaken the test should be this one:

https://github.com/simphony/simphony-common/blob/25e25f4bd29d55d96d030db68bcdcec98390611c/simphony/testing/abc_check_mesh.py#L135-L144

But for some reason is not running properly or is no using edges at all

PR #319, opened 2016-09-13
Problem: dynamically created data containers have bad base classes

mehdisadeghi:
Any class created with `create_data_container` helper function, which is supposed to be subclass of `DataContainer`, will not pass the following check:

```
DC = create_data_container()
dc = DC()
isinstance(dc, DataContainer) # False
```

Solution: add `__bases__` attribute to the created classes


itziakos:
I do not think that there is a need for the check isinstance(DataContainers). It is actuall wrong to do so since wrapper engines could have their own implementation of the DataContainer object. The important thing is to support the api for all CUDS objects as described in the docs.

Furthermore, manually manipulating `__base__` is beyond bad. If the need for an isinstance check is necessary the we need to have `ABCDataContainer` where compatible classes will `register` themselves and then isinstance(ABCDataContainer) will be true.

itziakos:
> Furthermore, manually manipulating `__base__` is beyond bad

Hmm.. I missed the fact that you are already creating a class at runtime. However, looking at the code `DataContainer` should be already part of the base classes since the [type] (https://docs.python.org/2.7/library/functions.html?highlight=type#type) function should have done it for you.

mehdisadeghi:
> DataContainer should be already part of the base classes since the type function should have done it for you

Strange. On my system it didn't work without explicitly setting `__bases__`. I'll double check if I'm missing something.

simphony/core/data_container.py
```diff
@@ -146,7 +146,8 @@ def create_data_container(restricted_keys,
     mapping = {key.name: key for key in restricted_keys}
 
     new_class = type(class_name, (DataContainer,),
-                     {'__doc__': DataContainer.__doc__,
+                     {'__bases__': (DataContainer),
```
itziakos: This also has to be a tuple `(DataContainer, )`

itziakos:
I think that there is a confusion. The `result` of the RestrictedContainer is a class type. when checking with `isnstance` you should check with `result()`

mehdisadeghi:
> you should check with result()

I checked with the instance. However, as you referenced the `type` doc, it should be already done by setting the bases tuple. I have definitely made a mistake during my debug. I'll probably close this PR.

itziakos:
> I have definitely made a mistake during my debug.

Yes I wanted to say that the instance, class type confusion could be done during the debug session.

mehdisadeghi:
This PR is misleading, I close it.

PR #321, opened 2016-10-13
Materials AbstractBaseClass and H5Materials implementation.

stefanoborini:
Introduces the Materials Abstract Base Class and the HDF5 implementation of Materials.


mehdisadeghi:
@stefanoborini we have to use the generated Material class instead of this one. I'll add more comments on the Slack channel.

stefanoborini:
According to @mehdisadeghi, Materials class is deprecated.

PR #322, opened 2016-10-13
PR: Feature: serialise CUDS model by Yaml script

tuopuu:
This PR adds a feature to save and load CUDS models from Yaml files. This is a work-in-progress version, but the users are encouraged to try it and test how it fits to different use-cases.

functions:
- save_CUDS(handle, model)
- load_CUDS(handle) (returns model)

### DIFF
--- a/simphony/io/serialisation.py
+++ b/simphony/io/serialisation.py
@@ -1,0 +1,291 @@
+import yaml
+import numpy
+
+from simphony.cuds.model import *
+from simphony.core.data_container import DataContainer
+from simphony.cuds.meta.cuds_component import CUDSComponent
+from simphony.core.keywords import KEYWORDS
+from simphony.cuds.meta import *
+from simphony.core.cuba import CUBA
+from collections import OrderedDict
+from importlib import import_module
+from simphony.cuds.meta.validation import to_camel_case
+
+_CUDSREFMAP = OrderedDict()
+
+
+def save_CUDS(handle, model):
+    """ Save CUDS model to a Yaml file
+
+    Parameters
+    ----------
+    handle: file handle
+        Yaml file where CUDS model is saved. File will be cleared.
+    model: CUDS model
+
+    """
+
+    handle.seek(0)
+    handle.truncate()
+
+    handle.write('- NAME:')
+    if model.name:
+        handle.write(' "'+model.name+'"')
+    else:
+        handle.write(' Null')
+    handle.write('\n\n')
+    handle.write('- DESCRIPTION:')
+    if model.description:
+        handle.write(' "'+model.description+'"')
+    else:
+        handle.write(' Null')
+    handle.write('\n')
+    #handle.write('# COMPONENTS')
+
+    # Construct a map of referenced objects
+    global _CUDSREFMAP
+    _CUDSREFMAP.clear()
+    for comp in model.iter(CUDSComponent):
+        # Add component to dictionary if it's not there already
+        if comp.uid not in _CUDSREFMAP.keys():
+            _CUDSREFMAP[comp.uid] = []
+        # Check if the component has CUDSComponents in data
+        for data in comp._data.values():
+            # either directly, or
+            if isinstance(data, CUDSComponent):
+                if data.uid not in _CUDSREFMAP.keys():
+                    _CUDSREFMAP[data.uid] = []
+                _CUDSREFMAP[data.uid] += [comp.uid]
+            # as an element of a list
+            if type(data) == list:
+                for elem in data:
+                    if isinstance(elem, CUDSComponent):
+                        if elem.uid not in _CUDSREFMAP.keys():
+                            _CUDSREFMAP[elem.uid] = []
+                        _CUDSREFMAP[elem.uid] += [comp.uid]
+
+    # Save CUDSComponents
+    for comp in model.iter(CUDSComponent):
+        if _CUDSREFMAP[comp.uid] is not 'saved':
+            stream = ''
+            stream = _CUDSComponent_to_yaml(comp, stream)
+            handle.write(stream)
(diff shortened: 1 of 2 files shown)


simphony/io/serialisation.py
```diff
@@ -0,0 +1,290 @@
...
+        # Find corresponding module and instantiate the class
+        mod = import_module('simphony.cuds.meta.%s' % cubatype.lower())
+        comp_class = getattr(mod, to_camel_case(cubatype))
+
+        # Must use (at least) two None parameters here, because currently
+        # there is no (easy) way to read required parameters for
+        # a class that is dynamically instantiated. A solution would be
+        # to define default values, e.g. 'None', for all parameters that
+        # do not have it defined yet.
+        comp_inst = comp_class(None, None)
```
mehdisadeghi: I think we have to agree on having no manatory arguements for instantiating meta classes. Nothing you can do here.

simphony/io/tests/test_serialisation.py
```diff
@@ -0,0 +1,147 @@
...
+        with closing(open(filename, 'r')) as handle:
+            CC = load_CUDS(handle)
+
+        self.assertEqual(CC.name, C.name)
+        self.assertEqual(CC.description, C.description)
+
+        for item in CC.iter(CUDSComponent):
+            self.assertEqual(item, None)
+
+    def test_save_CUDS_full(self):
```
mehdisadeghi: This test fails for me. I was not able to fix it in my first try.

simphony/io/tests/test_serialisation.py
```diff
@@ -0,0 +1,147 @@
...
+        filename = os.path.join(self.temp_dir, 'test_full.yml')
+        C = CUDS(name='full', description='fully randomized model')
+
+        for key in KEYWORDS.keys():
+            if key not in ['VERSION']:
```
mehdisadeghi: Why do we exclude `VERSION`?

simphony/io/tests/test_serialisation.py
```diff
@@ -0,0 +1,147 @@
...
+
+        # Iterate over components in the original model and check
+        # that they are present in the loaded model
+        for Citem in C.iter(CUDSComponent):
+            # Check items that have name parameter defined
+            if Citem.name is not None:
+                CCitem = CC.get(Citem.name)
+                self.assertEqual(CCitem.name, Citem.name)
+                for key in Citem.data:
+                    self.assertEqual(Citem.data[key], CCitem.data[key])
```
mehdisadeghi: While trying to fix the failed test I got weired errors this line. Like numpy array errors.

simphony/io/tests/test_serialisation.py
```diff
@@ -0,0 +1,147 @@
...
+                    return numpy.random.rand(shape[0])
+        elif dtype == str:
+            randstr = ''.join(random.choice(string.ascii_uppercase)
+                              for _ in range(shape[0]))
+            return randstr
+        elif dtype == bool:
+            return random.choice([True, False])
+
+    def _create_random_comp(self, key):
+        api = import_module('simphony.cuds.meta.api')
```
mehdisadeghi: Can't we simply import this on top of the module?

simphony/io/tests/test_serialisation.py
tuopuu: Version() needs four mandatory arguments and therefore we can't use ```comp_inst = comp_class(None,None)``` (see serialisation.py L242) to instantiate it. I could try to modify the L242 so that it would check there how many supported parameters the class has and instantiate it with corresponding number of Nones.

simphony/io/tests/test_serialisation.py
tuopuu: Yes, you're right. I'll fix that.

simphony/io/serialisation.py
kitchoi: Since the meta data specifies certain fields to be "required", it makes sense to have mandatory arguments, otherwise the "required" attribute is only a description in the yaml file that never gets executed or presented in any way in the class.  Alternatively, it should be quite simple to add a property called "required_keys" to the generated class.
Related to how `comp_class` should be instantiated: The arguments in the generated `__init__` may not be the same each time the class is generated, as the yaml file is loaded into a dictionary whose order is lost.  Therefore I would suggest *always* instantiating a class with keyword arguments.  i.e. `Box(name='box1')` instead of `Box('box1')` because one day the first argument might be something else that is not `name`

kitchoi: Could you try this?
```
init_kwargs = {required_parameters[0]: value1}
comp_inst = comp_class(**init_kwargs)
```
(and if required_parameters[0] is "CUBA.NAME", you probably want to convert it to "name")

tuopuu: Thanks @kitchoi. This is what I was looking for. I'll rewrite the code to instantiate classes that way, so we can get rid of randomly feeding None's to the objects.

mehdisadeghi:
Looks good to me. Please get rid of *WIP* tag whenever you are done.

tuopuu:
Thanks @mehdisadeghi . CUDS.data serialisation is still missing and I'm working on it right now. I'm writing data serialisation based on the modified data attribute which would be consistent with other objects having it defined. We can merge this feature after the data attribute is fixed.

tuopuu:
In fact, it's still an open question what should be stored in the CUDS.data DataContainer. Do we allow complicated CUDSComponents like Cfd() to be stored there, and what would be the use-case. Precisely, what's the distinction between a Cfd() object stored in CUDS.data and in CUDS._store?

I think we can merge this branch now and deal with the data attribute later, perhaps together with the incompletely tested saving of different CUDSComponents. I'm removing the WIP tag, so please, @mehdisadeghi go ahead and merge if you're happy with this branch.

PR #323, opened 2016-10-17
Problem: outdated meta classes are around

mehdisadeghi:
Solution: remove them.

This tiny PR removes three unused meta classes. Their respective metadata entries were changed and therefore these old ones are not necessary.


stefanoborini:
Which brings me to the fact that maybe we should not have them in the repo. Classes should be generated at develop, and certainly stored in the deployed package, but should not be physically committed to the repo.

Issue #190, opened 2015-07-14
Particle and Points are the same


roigcarlo:
I just detected that class Point in cuds.mesh and Particle in cuds.Particles are basically the same. Is this intentioned? I'm not aware of any plans to extend such classes with more functionalities.
Having only one would simplify the code, especially for tests.

The proposal is to create a class Point/Particle that will be used by both Mesh and Particles.

itziakos:
My personal opinion is that while these objects are equivalent at implementation level it would be confusing for the end user unless we find a neutral name. How about using the word `item`? Is that too generic.

SGGgarcia:
The name "Particle" is indeed a name that covers both particles and atoms concepts inside the atomistic world, so I think it will be difficult to find a name that can group different kind of entities in a conceptual level. Also `item` is too generic.
If the reason of this question is because the implementation is basically the same, maybe a good solution would be to have a private base class (not abstract) for both `Point` and `Particle` classes, and still keep these two classes, and keep coding and implementing for this base class instead. So a "mesh user" will use the Point class and a "particles user" will use the Particle class, and the implementation will remain private. Does this make sense?

itziakos:
> If the reason of this question is because the implementation is basically the same, maybe a good solution would be to have a private base class (not abstract) for both Point and Particle classes, and still keep these two classes, and keep coding and implementing for this base class instead. So a "mesh user" will use the Point class and a "particles user" will use the Particle class, and the implementation will remain private. Does this make sense?

I feel that having a common base class is more hassle that necessary since the Point and Partcle classes are very very simple and creating a class hierarchy for such simple objects is not going to help with testing (we still have to have separate code for that). 

I think that at the moment it is probably best to keep the current implementation.

PR #325, opened 2016-10-18
Fixed typo in test

stefanoborini:
Creating particles, while it should really create points. It is interesting to note that the tests don't pick the error.


mehdisadeghi:
> It is interesting to note that the tests don't pick the error.

Because `Particle` and `Point` have the exact same properties.

roigcarlo:
Not just the same properties, in fact there ARE the same. It was decided to let them exists as different objects because of the context in which they are used https://github.com/simphony/simphony-common/issues/190.

mehdisadeghi:
> Not just the same properties, in fact there ARE the same. 

Thanks for the historical reminder @roigcarlo !

PR #326, opened 2016-10-19
Generic interface for Mesh

stefanoborini:
This PR introduces the generic interface for datasets and applies it to the Mesh object, with minimal disruption to the internals and no disruptions to the old interface. 
We expect a loss of efficiency but that's the price to pay for not knowing the uid->type association. At the moment, it's probably not wise to implement optimization strategies to solve this issue, and will be part of a separate PR once we migrated all objects.


simphony/cuds/abc_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+            The object item.
+
+        Raises
+        ------
+        KeyError :
+            if any of the ids passed as parameters are not in the dataset.
+        """
+
+    @abstractmethod
+    def has(self, uid):  # pragma: no cover
```
mehdisadeghi: I suggest to implement `__contains__` instead of this.
https://docs.python.org/3/reference/datamodel.html#object.__contains__

simphony/cuds/abc_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+        item_type : CUDSItem
+            The CUDSItem enum of the type of the items to return the count of.
+
+        Returns
+        -------
+        True if the type is present, False otherwise.
+        """
+
+    @abstractmethod
+    def count_of(self, item_type):  # pragma: no cover
```
mehdisadeghi: Shell we implement `__len__` beside this? Of course it will return the whole number of elements inside the dataset.

simphony/cuds/abc_dataset.py
```diff
@@ -0,0 +1,174 @@
+# -*- coding: utf-8 -*-
+from abc import ABCMeta, abstractmethod
+
+
+class ABCDataset(object):
```
mehdisadeghi: DataSets must conform the the structore of `CUBA.DATA_SET`, i.e. they need `uid`, `name`, etc defined.

simphony/cuds/abc_mesh.py
```diff
@@ -1,21 +1,254 @@
...
+            the uids of the objects to be removed.
+
+        Raises
+        ------
+        KeyError :
+            If any object doesn't exist.
+        """
+        raise NotImplementedError("Remove is not implemented for Mesh")
+
+    def iter(self, uids=None, item_type=None):
```
mehdisadeghi: Docstring for `item_type` is missing

simphony/cuds/abc_dataset.py
mehdisadeghi: However, the same applies to all the other pythonic container methods. Perhaps it's better to leave this the subject of another PR. (And I see we have `has_something` methods there)

simphony/cuds/abc_dataset.py
stefanoborini: Ok then I probably used the wrong name. I just picked dataset but I am not sure about it.
What is the generic name for things like particles, mesh, and lattice?

simphony/cuds/abc_dataset.py
stefanoborini: well, if you do so, it's certainly possible, but at that point we should just reimplement the dict() interface... :)

simphony/cuds/abc_dataset.py
stefanoborini: That's possible, yes.

simphony/cuds/abc_dataset.py
mehdisadeghi: DataSet!

stefanoborini:
@mehdisadeghi please re-review.

simphony/cuds/abc_dataset.py
stefanoborini: Added.

simphony/cuds/abc_dataset.py
stefanoborini: Added

simphony/cuds/abc_mesh.py
stefanoborini: added

simphony/cuds/abc_dataset.py
mehdisadeghi: I think the name is fine. How about adding `meta.DataSet` as a parent to get the properties?

stefanoborini:
Some failures are due to the fact that we are now dependent on the DataSet. Some assumptions of the test are no longer valid, but they will be restored when we merge #335

I'll merge that branch into this one, but this mean that #335 must be merged first.

stefanoborini:
Another source of problem is that with the new hierarchy involving the generated base meta classes, the data object now contains None entries, that are converted to the string "None" when written on the hdf5 file with pytables.

stefanoborini:
The problem of None should be solved by setting appropriate defaults to strings, as from PR #343. In any case, the base classing will be done on a separate PR.

PR #327, opened 2016-10-19
Generic interface lattice

stefanoborini:
Built on top of #326. Please review that first.

Implements the generic interface for the Lattice object.


simphony/cuds/abc_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+        item_type : CUDSItem
+            The CUDSItem enum of the type of the items to return the count of.
+
+        Returns
+        -------
+        True if the type is present, False otherwise.
+        """
+
+    @abstractmethod
+    def count_of(self, item_type):  # pragma: no cover
```
mehdisadeghi: Please consider that when/if we define `Edge`, `Point`, `Particle`, `Bond`, `Face`, `Cell` and `Node` in the metadata (some of them are already defined), we have to get rid of `CUDSItem` enum.

stefanoborini: You mean as CUBA keywords? In that case, we'll have to change the implementation and the accepting value. I'd say we have to keep this as is, or we ruin the current interface, so it's just a step in that direction.

mehdisadeghi: I totally agree with you. However, we have to come back to this later.

PR #329, opened 2016-10-19
Command for meta build

stefanoborini:
Opening as a request for comments. 

This PR introduces a setup command to build the meta classes, keywords and cuba from the metadata repository. Once merged, we can get rid of the verbatim copied files, which are redundant (as they are self-generated), prone to error, and tend to leave debris.

However, I note that there has been a change in behavior with the new generator or current metainfo, so much that now tests fail. I've seen some changes in the classes, especially in shapes that used to be `[1]` and are now `[]` or `None`.

@mehdisadeghi does this make sense or is it a bug on this PR?

Please note that I had to work around some circular dependencies, so I might have missed something.


mehdisadeghi:
> shapes that used to be [1] and are now [] or None

I had to patch the generated modules in order to make tests pass. Namely, `keywords` and `cuby.py` modules. I've added `MATERIAL_ID` and set all `None` shapes to `[1]`.

stefanoborini:
@mehdisadeghi Are you saying that the automatic generator produces something that is broken, and we need to patch the result manually?

stefanoborini:
@mehdisadeghi if so, this needs fixing.

mehdisadeghi:
@stefanoborini please review the recent changes.

PR #335, opened 2016-10-24
Copying data container

stefanoborini:
Performs copy of the data container at every interaction, as agreed with @mehdisadeghi


scripts/generate.py
```diff
@@ -910,6 +927,7 @@ def meta_class(yaml_file, out_path, overwrite):
     with make_temporary_directory() as temp_dir:
 
         for key, class_data in yml_data['CUDS_KEYS'].items():
+            print("cuds keys ", key, class_data)
```
stefanoborini: print

stefanoborini:
@mehdisadeghi ready for review

scripts/generate.py
```diff
@@ -641,23 +650,25 @@ def data(self):
...
+        return retvalue''')
 
         self.methods.append('''
     @data.setter
     def data(self, new_data):
-        if isinstance(new_data, DataContainer):
-            self._data = new_data
-        else:
-            self._data = DataContainer(new_data)''')
+        data = DataContainer.new_with_restricted_keys(
```
mehdisadeghi: The keys shall remain flexible. All `CUBA` keys are accepted.

scripts/generate.py
```diff
@@ -819,8 +830,10 @@ def generate_initializer(self, file_out):
         '''
         # __init__ keyword arguments
         kwargs = []
-        for key, content in chain(self.inherited_optional.items(),
-                                  self.optional_user_defined.items()):
+        for key, content in chain(
+                self.optional_user_defined.items(),
+                self.inherited_optional.items(),
+        ):
```
kitchoi: I see that this change in order has resulted in the changes in the generated __init__ signatures.  Is this intentional?
Just an observation: I suppose we cannot fix the signature such that future additional optional arguments would only be added at the end of a signature.  Somewhere (in the doc maybe) the users should be advised to use named arguments when instantiating meta classes.

stefanoborini:
@kitchoi yes, it's intentional, because it was currently inconsistent.

kitchoi:
@sbo Ok!  Thanks.

scripts/generate.py
mehdisadeghi: It is still valid. We want to leave it flexible, however. To the extend of CUBA keys.

mehdisadeghi: It is rendered useless. People used to like it. Deprecate it if you like.

stefanoborini:
We decided to copy everything, ignoring the supported parameters.

stefanoborini:
Also, see issue #341

mehdisadeghi:
Everything looks good to me!

mehdisadeghi:
Related PR #326

PR #337, opened 2016-10-26
Problem: travis config uses system-wide python packages

mehdisadeghi:
Solution: tell it not to use it

Moreover run tests with `py.test -v` which is smarter and produces nicely formatted output. It is installed by default in travis containers.


mehdisadeghi:
@stefanoborini this PR should solve the problem.

PR #343, opened 2016-11-02
Ensures strings are initialized with the proper empty default

stefanoborini:
Strings were not initialized properly when they were "plain strings" with no "CUBA" component. This resulted in NAME and DESCRIPTION be set to None, which introduces problems with HDF5 storage.


mehdisadeghi:
Related PR #326

PR #346, opened 2016-11-04
Internalized data

stefanoborini:
Internalizes the data member without explicit specification. This change needs the removal of "data" from the yaml file. Also fixes floating point misrepresentation. See PR simphony/simphony-metadata#56 and simphony/simphony-metadata#58


tuopuu:
While at it, use of CUDS.data attribute (in model.py) could be made consistent with the CUDSComponents. Right now there is a conflict with CUDS.name and CUDS.description as they are not stored in the CUDS.data attribute. User can also add a different description/name in the data attribute which makes no sense. It should be enough to have only one name/description attribute in the model.

mehdisadeghi:
> While at it, use of CUDS.data attribute (in model.py) could be made consistent with the CUDSComponents. 

Good point. I'll make a couple of PRs which will include that.

mehdisadeghi:
@stefanoborini since https://github.com/simphony/simphony-metadata/pull/56 is closed let's see how we can proceed with this PR.

stefanoborini:
@mehdisadeghi please review #350 first. It's unrelated, but I want to merge that one and #348 to reduce chance for conflicts.

PR #347, opened 2016-11-07
Problem: the generated cuba enum is unsorted, hence hard to read

mehdisadeghi:
Solution: sort the enum in generation time


mehdisadeghi:
@stefanoborini please review

PR #349, opened 2016-11-08
Problem: presence of `data` in init method is confusing

mehdisadeghi:
Solution: skip `data` property upon code generation


stefanoborini:
Please don't. I'll do it in my branch.

PR #350, opened 2016-11-09
Specify metadata repo tag to checkout

stefanoborini:
We constantly see generated classes change as we rebuild against the most recent metadata repo, without any indication of which repository tag/commit we are actually using. This change allows us to keep track of what we are building against.


mehdisadeghi:
Looks good to me.

PR #352, opened 2016-11-11
Problem: meta is outdated

mehdisadeghi:
Solution: regenerate based on the latest metadata commit


mehdisadeghi:
@stefanoborini just applied your changes

PR #353, opened 2016-11-11
Problem: CUDSItem enum is duplicate

mehdisadeghi:
Solution: replace it with CUBA everywhere

Backgroud: all the keys in CUDSItem enum are now defined in the metadata and have their own CUBA keys.


mehdisadeghi:
@stefanoborini please review.

PR #362, opened 2016-12-12
Problem: `data` property should be a reference

mehdisadeghi:
Solution: revert corresponding change made in #335 

@ahashibon @stefanoborini I reverted a changed which was my bad.

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -10,3 +10,4 @@ coverage
 flake8
 numpy >= 1.4.1
 hypothesis
+yapf
--- a/scripts/generate.py
+++ b/scripts/generate.py
@@ -646,7 +646,7 @@ class CodeGenerator(object):
         self.methods.append('''
     @property
     def data(self):
-        return DataContainer(self._data)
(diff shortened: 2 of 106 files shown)


mehdisadeghi:
@stefanoborini @ahashibon The regenerate solves also the `FACE` change (which is removed from cuba.yml)

ahashibon:
Thanks @mehdisadeghi, looks good to me.

stefanoborini:
Ok, now I am confused. Didn't we agree that we had to return a copy?

stefanoborini:
Ok, this PR is doing unrelated things so I will do the following:

1. update against a new version of simphony-metadata, extracting the relevant bits.
2. if green, merge it back here.
3. merge the yapf patch I have already.

mehdisadeghi:
@stefanoborini This PR is desired since it provides the expected behavior. 

BTW, I realized even simphony v3.0 has been inconsistent regarding `data` property. Some classes such as `Particle` treat it as reference and some others like `Particles` treat it as copy.

PR #367, opened 2017-01-06
Fixing typo error in decode_shape

stefanoborini:
This is not really important as the code will go away, but it was there
and I don't like it around


stefanoborini:
Incorrect. I misread the regexp.

PR #376, opened 2017-01-24
Problem: model.py causes confusion

mehdisadeghi:
Solution: rename it to cuds.py

I noticed an error regarding uid in yaml serialization and solved that as well.


stefanoborini:
are there any wrappers using these classes?

mehdisadeghi:
> From our discussion there should not be, but just in case.

AFAIK there should not be. Moreover, this does not break imports from `simphony.cuds`.

PR #377, opened 2017-01-24
Regenerate meta to add CUDS

mehdisadeghi:
@ahashibon This PR adds `CUBA.CUDS` entry and its corresponding `CUDS` class.


ahashibon:
@mehdisadeghi looks good. Please merge.

PR #378, opened 2017-01-25
Problem: datasets do not have uids

mehdisadeghi:
Solution: add uid to datasets

including Particles, Lattice and Mesh.


ahashibon:
Looks good.

simphony/core/keywords.py
```diff
@@ -1012,6 +1012,13 @@
         shape=[1],
         length=None,
         dtype=None),
+    'CUDS': Keyword(
+        name='CUDS',
+        definition='CUDS Container',  # noqa
```
stefanoborini: Please explain better what this is.

stefanoborini: In the metadata, of course. This is generated code.

mehdisadeghi: Sure. We'll do it in metadata repository.

PR #386, opened 2017-02-10
Pinned to 0.1.0 metaparser.

stefanoborini:
0.2.0.dev0 changes the interface, so for now we pin against 0.1.0


stefanoborini:
I need the extended API. Closing

PR #394, opened 2017-02-16
Problem: some wrappers fail to load on Simulation import

mehdisadeghi:
Solution: make sure engines are loaded before any other import.

This prevent cyclic imports.


stefanoborini:
do you have an actual error message for this? because i don't see the potential for a cyclic import

mehdisadeghi:
@stefanoborini try running nve_example in lammps without importing simphony.engine in the script, you will see the error. Actually it is bad that we implicitely load the engines, but it is a story for another day.

PR #395, opened 2017-03-07
Fix datacontainer particles, mesh and lattice datasets

ahashibon:
DS.data should return the data container, and not a copy. This is consistent with dataset elements and metadata generated classes (cuds_item). 
@mehdisadeghi please review.

### DIFF
--- a/simphony/cuds/lattice.py
+++ b/simphony/cuds/lattice.py
@@ -83,7 +83,7 @@ class Lattice(ABCLattice):
 
     @property
     def data(self):
-        return DataContainer(self._data)
+        return self._data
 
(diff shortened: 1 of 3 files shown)


stefanoborini:
Good for me, provided that the client code is aware of this change

ahashibon:
@stefanoborini can we merge this?

stefanoborini:
@ahashibon I just verified that the same change needs to be done in the generated classes. This first requires a change in the generator, then a rebuild of the meta classes in this PR.

stefanoborini:
@ahashibon correction, it's already there. Merging.

PR #396, opened 2017-03-10
Revert relocation of validation support code to the module itself

stefanoborini:
The validation module must be self-contained.

Reason is that in principle, we should not have scripts in here, and in fact on master, scripts then further imports simphony_metaparser, which is a development package and thus may not be present in a non-dev installation.


stefanoborini:
Closed for incorrect branch.

PR #425, opened 2017-06-30
Edm build infrastructure

stefanoborini:
Implements the edm build infrastructure to deploy the package on edm server


dpinte:
@stefanoborini LGTM

